PR #20, opened 2024-05-31
dynamicMetadata preset

jeevitha011:
- `dynamicMetadata: true` can be set only for production.
- By default it is false.
- Supported for Node.js and Java.

To test:
- Run `cds build --production`
- gen/srv contains metadata files
- With `dynamicMetadata` set by default or to false, deploy application
- Static files are read when metadata endpoints are hit


RoshniNaveenaS (commented):
Could be merged along with java ord support

PR #23, opened 2024-06-03
split package based on policyLevel

s-shreyas-s:
Split package array based on policyLevel value


RoshniNaveenaS (approved):
Tested and looks good

PR #29, opened 2024-06-18
adding gif

RoshniNaveenaS:


README.md
```diff
@@ -38,6 +38,8 @@ cds compile <path to srv folder> --to ord [-o] [destinationFilePath]
 1) Run `cds watch` in the application's root.
 2) Check the following relative paths for ORD information - `/.well-known/open-resource-discovery` , `/open-resource-discovery/v1/documents/1`.
 
+![](ord.gif)
```
jeevitha011, line 41: you should use image tag like [this](https://github.com/cap-js/notifications/blob/main/README.md?plain=1#L114)

README.md
```diff
@@ -38,6 +38,8 @@ cds compile <path to srv folder> --to ord [-o] [destinationFilePath]
...
+
```
jeevitha011, line 42: the ord endpoint url is not visible in the gif

jeevitha011, line 42: - file size is too big and should be compressed as it takes time to load
- gif should be sped up as it is very slow

jeevitha011, line 42: 'site cant be reached' part of the gif should be edited out

README.md
```diff
@@ -39,6 +39,9 @@ cds compile <path to srv folder> --to ord [-o] [destinationFilePath]
 2) Check the following relative paths for ORD information - `/.well-known/open-resource-discovery` , `/open-resource-discovery/v1/documents/1`.
 
 
+<img width="1300" alt="Sample Application Demo" style="border-radius:0.5rem;" src="ord.gif">
```
jeevitha011, line 42: ord.gif should be in a separate folder called asset/etc

RoshniNaveenaS:
Have added the updated gif

README.md
RoshniNaveenaS: made the changes

README.md
RoshniNaveenaS: now it is

RoshniNaveenaS: have increased the speed

RoshniNaveenaS: have removed

README.md
muskansethi1, line 42: update src, the gif is not loading in the readme

README.md
```diff
@@ -39,6 +39,9 @@ cds compile <path to srv folder> --to ord [-o] [destinationFilePath]
...
+
+
```
muskansethi1, line 44: For me, the gif goes to white and gets stuck on it after one point, is it supposed to be that way?

README.md
RoshniNaveenaS: made changes

README.md
RoshniNaveenaS: have updated that can u check again

muskansethi1:
personal opinion - we could use an incognito browser window, so that the gif doesn't have the bookmarks bar or any other info. It would look cleaner, and avoid any personal info being displayed

RoshniNaveenaS:
> imo - we could use an incognito browser window, so that the gif doesn't have the bookmarks bar or any other info. It would look cleaner, and avoid any personal info being displayed

made changew

muskansethi1 (approved):

PR #36, opened 2024-07-11
Minor fixes

RoshniNaveenaS:


shubho5:
General comment:
In the future, please avoid using global as a function parameter or passing it as a value of the parameter. It doesn't serve the purpose of global except in some cases. In your code, I think it can be avoidable.

shubho5 (approved):
Approved.

Issue #38, opened 2024-07-25
ORD Metadata changes


RamIndia:
Please find attached the ord documents we walked through in our review: 
- Reference CAP application: https://github.com/SAP-samples/cloud-cap-samples/tree/main/bookstore
- [1.ord.json](https://github.com/user-attachments/files/16372814/1.ord.json) is the one generated today by cds compile index.cds --to ord in cap/samples/bookstore
- [2.ord.json](https://github.com/user-attachments/files/16372820/2.ord.json) is the same with corrections 
- → compare both in VSCode, e.g. with code --diff 1.ord.json 2.ord.json

Here are the changes decided for the ORD document : 

- [x] Don’t add a suffix if I already specified a **.json** suffix in the **-o / —dest** parameter
- [x] Use **.ord.json** as file suffix instead of .ord if not 
- [x] Always add  ```"$schema": "https://sap.github.io/open-resource-discovery/spec-v1/interfaces/Document.schema.json"```
- [x] Never use CSN's **.namespace** property but always just use the fully-qualified names from **.definitions** 
- [x] ORD namespace (used as prefix in several ORD IDs) = `package.name.replace(/@/,'').replace(/\//g,’')`, unless explicitly configured in cds.env.ord.namespace

**products** 

- [x] `"ordId": "customer:product:capire.bookstore:”`→ i.e. package.name.replace(/@/,'').replace(/\//g,'.')
- [x] `"vendor": "customer:vendor:customer:” `
- [x] `"title": “capire bookstore”`→ i.e. package.name.replace(/@/,'').replace(/\//g,'.')
- [x] `"shortDescription": “capire bookstore"` → i.e. package.name.replace(/@/,'').replace(/\//g,' ')

**groups** 

- [x] `"title": "Catalog Service”` → i.e. without appending “ Title”
- [x] The groups for events shall be services, not the individual events → but then the last two entries in the bookstore sample are duplicates

**packages**

- [x] → the namespace was wrong and the package name shall be `package.name.replace(/@/,'').replace(/\//g,'.')`
- [x] `"partOfProducts": [ "customer:product:capire.bookstore:” ]` → has to be the same as the product name in the products section
- [x] `"vendor": "customer:vendor:customer:”` → has to be the same as the vendor in the products section
- [x] `"title": "capire bookshop”` → take what is there, don’t add extra text like “sample title for"
- [x] `“shortDescription"` → take what is there, don’t add extra text like “Here’s the short description for ..."
[...]

PR #42, opened 2024-08-21
feat: Use package name for product ORD IDs.

ductaily:
Addressing 

- ORD namespace (used as prefix in several ORD IDs) = `package.name.replace(/@/,'').replace(/\//g,’')`, unless explicitly configured in cds.env.ord.namespace
- **Products:** 
    - "ordId": "customer:product:capire.bookstore:”→ i.e. `package.name.replace(/@/,'').replace(/\//g,'.')`
    - `"vendor": "customer:vendor:customer:”`
    - "title": “capire bookstore”→ i.e. `package.name.replace(/@/,'').replace(/\//g,' ')`
    - "shortDescription": “capire bookstore" → i.e. `package.name.replace(/@/,'').replace(/\//g,' ')` - Description for + "package name"

from https://github.com/cap-js/ord/issues/38

PR #47, opened 2024-08-28
feat: Address issues with packages

ductaily:
- [x] → the namespace was wrong and the package name shall be `package.name.replace(/@/,'').replace(/\//g,'.')`
- [x] `"partOfProducts": [ "customer:product:capire.bookstore:” ]` → has to be the same as the product name in the products section
- [x] `"vendor": "customer:vendor:customer:”` → has to be the same as the vendor in the products section
- [x] `"title": "capire bookshop”` → take what is there, don’t add extra text like “sample title for"
- [x] `“shortDescription"` → take what is there, don’t add extra text like “Here’s the short description for ..."
- [x] `“description"` → take what is there, don’t add extra text like “Here’s the short description for ..."

In https://github.com/cap-js/ord/issues/38

### DIFF
--- a/lib/defaults.js
+++ b/lib/defaults.js
@@ -13,6 +13,8 @@ const nameWithSpaces = (name) => {
   return regexWithRemoval(name.charAt(0)) + name.slice(1, name.length).replace(/[^a-zA-Z0-9]/g,' ');
 }
 
+const defaultProductOrdId = (name) => `customer:product:${nameWithDot(name)}:`
+
 /**
  * Module containing default configuration for ORD Document.
  * @module defaults
@@ -24,7 +26,7 @@ module.exports = {
   description: "this is an application description",
   products: (name) => [
     { 
-      ordId: `customer:product:${nameWithDot(name)}:`,
+      ordId: defaultProductOrdId(name),
       title: nameWithSpaces(name),
       shortDescription: "Description for " +  nameWithSpaces(name),
       vendor: "customer:vendor:customer:",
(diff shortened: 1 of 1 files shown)


lib/defaults.js
```diff
@@ -35,12 +37,12 @@ module.exports = {
     function createPackage(name, tag) {
       return {
         ordId: `${regexWithRemoval(name)}:package:${capNamespace}${tag}`,
-        title: `sample title for ${regexWithRemoval(name)}`,
-        shortDescription: "Here is the shortDescription for packages",
-        description: "Here is the description for packages",
+        title: nameWithSpaces(name),
+        shortDescription: "Description for " + nameWithSpaces(name),
```
Fannon, line 41:
```suggestion
        shortDescription: "Short description for " + nameWithSpaces(name),
```

ductaily:
If I understood correctly the first bullet point was addressed in the PR for the Products
`→ the namespace was wrong and the package name shall be package.name.replace(/@/,'').replace(/\//g,'.')`

Is my understanding correct? @RoshniNaveenaS 

CC: @aramovic79 @Fannon

lib/defaults.js
Fannon: Maybe if we explicitly mention the prefixes like that, it motivates people to look into where to change it, once they see it published / exposed somewhere. Then it would be good to see that one is the short description and the other is the (full) description.

lib/defaults.js
```diff
@@ -35,12 +37,12 @@ module.exports = {
...
+        description: "Description for " + nameWithSpaces(name),
         version: "1.0.0",
-        partOfProducts: [`customer:product:${nameWithDot(name)}:`],
-        vendor: "customer:vendor:SAP:"
+        partOfProducts: [defaultProductOrdId(name)],
+        vendor: "customer:vendor:customer:"
```
Fannon, line 45:
```suggestion
        vendor: "customer:vendor:Customer:"
```

This is the correct value. It's a reserved Customer vendor, but it should also be documented on the public ORD spec. I couldn't find this on first glance, so I'll add it later.

RoshniNaveenaS:
> d in the PR for the Prod

Yes we had the fix in [this PR](https://github.com/cap-js/ord/pull/42/files)

ductaily:
> > d in the PR for the Prod
> 
> Yes we had the fix in [this PR](https://github.com/cap-js/ord/pull/42/files)

Thank you for the confirmation.

Fannon (approved):

RoshniNaveenaS (approved):
Looks good to me

PR #51, opened 2024-08-28
Feat: upgrade cds version to 8

zongqichen:


package.json
```diff
@@ -19,7 +19,7 @@
     "eslint": "^8"
   },
   "peerDependencies": {
-    "@sap/cds": "7.9.4",
+    "@sap/cds": "^8",
```
ductaily, line 22: I would specify a full version instead of just the major version.

RoshniNaveenaS (approved):

Issue #50, opened 2024-08-28
Feat: upgrade @sap/cds version to 8


zongqichen:
There is @sap/cds major version release since Jul, you can have a look here: [CAP Node.js release](https://cap.cloud.sap/docs/releases/schedule#cap-node-js). It brings some new important features, for example, (containerize)[https://cap.cloud.sap/docs/guides/deployment/to-kyma#build-images] for building docker image.

PR #53, opened 2024-08-29
feat: Skip generation of consumption bundles

ductaily:
**consumptionBundles**

- [x]  skip (do not include)

**Test**
- [x] add jest test (should be extended)
- [x] add GitHub workflow to execute tests on PR against `main` or commit to `main`


jest.config.js
```diff
@@ -0,0 +1,6 @@
+const config = {
+    testTimeout: 42222,
```
Fannon, line 2: Why such a huge number? Please use something more reasonable. If it's in ms, something like:

```suggestion
    testTimeout: 4000,
```

ductaily: Right :) I did not think about it. Thank you for the suggestion.

tests/ord.test.js
```diff
@@ -0,0 +1,137 @@
+const cds = require("@sap/cds");
```
zongqichen, line 1: Good to see, we have unittest already. I would suggest that puts the unittest file into a folder named "unittest". Because, we need to have integration test as well.

package.json
```diff
@@ -13,15 +13,17 @@
...
+    "eslint": "^8",
+    "jest": "^29.7.0"
   },
   "peerDependencies": {
-    "@sap/cds": "7.9.4",
-    "@sap/cds-compiler" : "5.0.6",
     "@cap-js/asyncapi": "^1.0.0",
-    "@cap-js/openapi": "^1.0.2"
+    "@cap-js/openapi": "^1.0.2",
+    "@sap/cds": "7.9.4",
```
zongqichen, line 26: Could we upgrade cds to 8 version, there is an issue existing: https://github.com/cap-js/ord/issues/50

jest.config.js
Fannon: Having a lower number here also has the nice benefits that you could argue that slow tests indicate a performance bug, that needs fixing ;)

And if something is timing out, then CI/CD will fail faster.

tests/ord.test.js
Fannon: I think jest has some conventions like having the folder name of `__tests__` (or similar). Maybe we can distinguish the different categories of tests by adding tags or prefixing the describe block with the kind of test it is. 
Then it should be possible to filter and only run a specific kind of test.

tests/ord.test.js
```diff
@@ -0,0 +1,137 @@
...
+describe("Tests for default ORD document", () => {
+  let csn;
+
+  beforeAll(async () => {
+    csn = await cds.load(path.join(__dirname, "bookshop", "srv"));
+  });
+
+  test("Successfully create ORD Documents with defaults", () => {
+    const document = ord(csn);
+    expect(document).toStrictEqual({
```
Fannon, line 14: I would not copy the entire output of the test in here. If you want to do a simple check that the result does not change without you noticing, you can use a jest snapshot instead. See https://jestjs.io/docs/snapshot-testing

ductaily: I wasn't aware of this feature. Very cool. I changed the test to use a snapshot.

tests/ord.test.js
ductaily: I renamed with the folder to `__tests__` since this is the convention.
I would agree with @Fannon to use tags or prefixes in describe blocks instead.

Fannon (approved):

tests/ord.test.js
zongqichen: > I think jest has some conventions like having the folder name of `__tests__` (or similar). Maybe we can distinguish the different categories of tests by adding tags or prefixing the describe block with the kind of test it is. 
> 
> Then it should be possible to filter and only run a specific kind of test. 

In this pr, without unit test folder would fine. But I would prefer fully distinguishing unit test, integration test and e2e test. We can use tag to call some reusable tests, however we can put them into common utils. In principle, those three type of tests should have clear different logic and targets. But we can achieve it in following pr.

zongqichen (approved):

tests/ord.test.js
Fannon: Agree. There's also a 4th kind of test in my experience: The API tests, which use a request library to call the REST APIs from the outside. Or is this what you already had in mind with e2e? 
With e2e, I'm always thinking of UI testing, like cypress :)

PR #54, opened 2024-09-02
Feat: Add Unittests

zongqichen:


__tests__/unittest/templates.test.js
```diff
@@ -0,0 +1,197 @@
...
+                }
+            ]
+            expect(templates.checkEntityFunctionAction(linkedModel, global)).toEqual(testResult);
+        });
+
+        it('should return actions', () => {
+        });
+
+        it('should return function', () => {
+        });
```
ductaily, line 45: Please remove empty tests if there are no tests.

__tests__/unittest/templates.test.js
```diff
@@ -0,0 +1,197 @@
...
+            const testString = 'customer.testNamespace.123';
+            expect(templates.fReplaceSpecialCharacters(testString)).toEqual('customer.testNamespace123');
+        });
+
+        it('should return replaced dash', () => {
+            const testString = 'customer.testNamespace-123';
+            expect(templates.fReplaceSpecialCharacters(testString)).toEqual('customer.testNamespace123');
+        });
+
+        // TODO: is that correct?
```
ductaily, line 60: Just a proposal: Let's remove the TODO comments out and move them in a GitHub issue instead

ductaily (changes requested):

__tests__/unittest/templates.test.js
zongqichen: Good point, I left here as placeholder. The test case didn't cover every conditions. But we need to keep main branch clean enough. I will clean up all placeholder and TODO comments

__tests__/unittest/templates.test.js
zongqichen: Yes, will do

__tests__/unittest/defaults.test.js
```diff
@@ -0,0 +1,147 @@
+const defaults = require('../../lib/defaults');
+
+describe('defaults', () => {
+    describe('$schema', () => {
+        const test$schemaDefaultValue = 'https://sap.github.io/open-resource-discovery/spec-v1/interfaces/Document.schema.json'; 
+        it('should return default value', () => {
+            expect(defaults.$schema).toEqual(test$schemaDefaultValue);
```
Fannon, line 7: Maybe let's have a meeting on this how we generally want to write unit-tests.

I think a test like this is not giving us more value than a snapshot test (that we already have), but is more cumbersome to maintain. For the "isEqual" tests, I think we can just rely on the Snapshot test to ensure that we don't accidentally change something without first reviewing and accepting the change. 

For the real unit-tests, I would prefer to have them test more logical things, like that there's no "undefined" string somewhere in a generated ORD document (we had such issues creep in before), or that the ORD ID of the package assignment can be resolved within the ORD document. 

We could test equals assumptions where it's really critical or has special meaning, like that the default policy level of default is "sap:base:v1". 

@zongqichen , @ductaily , @aramovic79

zongqichen: Hi @Fannon, thank you for valuable comments.  We need a meeting discussing about unittest and code style with team agreement. From my understanding and experiences, uniitest should be simple, quick, fragile, readable and understandable. If the new codes break the unittest, we have a chance to think twice, either the issue from codes or tests. This is the front safeguard for us. And maintaining unittest should be as same important as production code, we should not worry about maintaining test.

The "snapshot test" you mentioned is more like integration test for me. The purpose of integration test is to watch how it looks like in general at the end with different environment settings or complex scenarios, which unittest can not accomplish. 

Comparing with snapshot test, I feel the test you mentioned has more values in different scenarios:
- The fine-grained unittest is easier and quicker to locate the issue. And during development, we can also run the fine-grained test independently, I think it's better to debug as well.
- Later, I want to clean and refactor the code for fixing naming issue, reducing function complexity and duplicable. The fine-grained unittest can support this process.
- I have same feeling with you. This PR is more like uniitest proposal and skeleton. We need to improve it with more test cases and more logics. Later, I'm pretty sure, we can reuse snapshot to reduce the line of codes. Back to this test, the target function is simply enough, I assume more flexible and testcases will be enough.

btw:
I found the bug undefined during writing unittest: https://github.com/cap-js/ord/blob/b914cfa0e02f979cb25871beae97d7705bf583ea/__tests__/unittest/templates.test.js#L20, it's missing error handling or default condition.

Fannon (approved):

ductaily (approved):

PR #55, opened 2024-09-02
fix: event resources generation

ductaily:
Address issues from https://github.com/cap-js/ord/issues/38

eventResources

- [x] The entries are service-level, not event-level
- [x] IMPORTANT: test with more than one event declared per service → should end up in one entry, not many

### DIFF
--- a/__tests__/__snapshots__/ord.test.js.snap
+++ b/__tests__/__snapshots__/ord.test.js.snap
@@ -1,0 +1,126 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Tests for default ORD document Successfully create ORD Documents with defaults 1`] = `
+{
+  "$schema": "https://sap.github.io/open-resource-discovery/spec-v1/interfaces/Document.schema.json",
+  "apiResources": [
+    {
+      "apiProtocol": "odata-v4",
+      "description": "Here we have the description for AdminService",
+      "entityTypeMappings": [
+        {
+          "entityTypeTargets": [],
+        },
+      ],
+      "entryPoints": [
+        "/odata/v4/admin",
+      ],
+      "extensible": {
+        "supported": "no",
+      },
+      "ordId": "capjs.ord:apiResource:undefined.AdminService:v1",
+      "partOfGroups": [
+        "sap.cds:service:capjs.ord:undefined.AdminService",
+      ],
+      "partOfPackage": undefined,
+      "releaseStatus": "active",
+      "resourceDefinitions": [
+        {
+          "accessStrategies": [
+            {
+              "type": "open",
+            },
+          ],
+          "mediaType": "application/json",
+          "type": "openapi-v3",
+          "url": "/.well-known/open-resource-discovery/v1/api-metadata/AdminService.oas3.json",
+        },
+        {
+          "accessStrategies": [
+            {
+              "type": "open",
+            },
+          ],
+          "mediaType": "application/xml",
+          "type": "edmx",
+          "url": "/.well-known/open-resource-discovery/v1/api-metadata/AdminService.edmx",
+        },
+      ],
+      "shortDescription": "Here we have the shortDescription for AdminService",
+      "title": "The service is for AdminService",
+      "version": "1.0.0",
+      "visibility": "public",
+    },
+    {
+      "apiProtocol": "odata-v4",
+      "description": "Here we have the description for CatalogService",
+      "entityTypeMappings": [
+        {
+          "entityTypeTargets": [],
+        },
+      ],
+      "entryPoints": [
+        "/browse",
+      ],
+      "extensible": {
+        "supported": "no",
+      },
+      "ordId": "capjs.ord:apiResource:undefined.CatalogService:v1",
+      "partOfGroups": [
+        "sap.cds:service:capjs.ord:undefined.CatalogService",
+      ],
+      "partOfPackage": undefined,
+      "releaseStatus": "active",
+      "resourceDefinitions": [
+        {
+          "accessStrategies": [
+            {
+              "type": "open",
+            },
+          ],
+          "mediaType": "application/json",
+          "type": "openapi-v3",
+          "url": "/.well-known/open-resource-discovery/v1/api-metadata/CatalogService.oas3.json",
+        },
+        {
+          "accessStrategies": [
+            {
+              "type": "open",
+            },
+          ],
+          "mediaType": "application/xml",
+          "type": "edmx",
+          "url": "/.well-known/open-resource-discovery/v1/api-metadata/CatalogService.edmx",
+        },
+      ],
+      "shortDescription": "Here we have the shortDescription for CatalogService",
+      "title": "The service is for CatalogService",
+      "version": "1.0.0",
+      "visibility": "public",
+    },
+  ],
+  "description": "this is an application description",
+  "groups": [
+    {
+      "groupId": "sap.cds:service:capjs.ord:undefined.AdminService",
+      "groupTypeId": "sap.cds:service",
+      "title": "Admin Service Title",
+    },
+    {
+      "groupId": "sap.cds:service:capjs.ord:undefined.CatalogService",
+      "groupTypeId": "sap.cds:service",
+      "title": "Catalog Service Title",
+    },
+  ],
+  "openResourceDiscovery": "1.9",
+  "policyLevel": "none",
+  "products": [
+    {
+      "ordId": "customer:product:cap.js.ord:",
+      "shortDescription": "Description for cap js ord",
+      "title": "cap js ord",
+      "vendor": "customer:vendor:customer:",
+    },
+  ],
+}
+`;
--- a/__tests__/__snapshots__/ord.test.js.snap
+++ b/__tests__/__snapshots__/ord.test.js.snap
@@ -22,7 +22,7 @@ exports[`Tests for default ORD document Successfully create ORD Documents with d
       "partOfGroups": [
         "sap.cds:service:capjs.ord:undefined.AdminService",
       ],
-      "partOfPackage": undefined,
+      "partOfPackage": "capjsord:package:undefined:v1",
       "releaseStatus": "active",
       "resourceDefinitions": [
         {
@@ -69,7 +69,7 @@ exports[`Tests for default ORD document Successfully create ORD Documents with d
       "partOfGroups": [
         "sap.cds:service:capjs.ord:undefined.CatalogService",
       ],
-      "partOfPackage": undefined,
+      "partOfPackage": "capjsord:package:undefined:v1",
       "releaseStatus": "active",
       "resourceDefinitions": [
         {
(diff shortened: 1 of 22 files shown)


ductaily:
I extended service definitions of bookshop in the test:

```js
using { sap.capire.bookshop as my } from '../db/schema';
service AdminService @(requires:'authenticated-user') { 
  entity Books as projection on my.Books;
  entity Authors as projection on my.Authors;

  event BookCreated : {
    ID    : Integer;
    title : String @title: 'Title';
  };

  event BookDeleted : {
    ID    : Integer;
  };

  event BookUpdated : {
    ID    : Integer;
    title : String @title: 'Title';
  }
}
```

```js
using { sap.capire.bookshop as my } from '../db/schema';
service CatalogService @(path:'/browse') { 

  @readonly entity Books as select from my.Books {*,
    author.name as author
  } excluding { createdBy, modifiedBy };

  @requires: 'authenticated-user'
  action submitOrder (book: Books:ID, quantity: Integer);
  
  event BookCreated : {
      ID    : Integer;
    title : String @title: 'Title';
  };

  event BookDeleted : {
    ID    : Integer;
  };

  event BookUpdated : {
    ID    : Integer;
    title : String @title: 'Title';
  }
}

```

Resulting ORD Document:

```json5
{
  "apiResources": [
     //...
  ],
  "description": "this is an application description",
  "eventResources": [
    {
      "description": "This is an example event catalog that contains only a partial ODM cap-js-ord V1 event",
      "extensible": {
        "supported": "no",
      },
      "ordId": "capjs.ord:eventResource:undefined.AdminService:v1",
      "partOfGroups": [
        "sap.cds:service:capjs.ord:undefined.AdminService",
      ],
      "partOfPackage": "capjsord:package:undefined:v1",
      "releaseStatus": "beta",
      "resourceDefinitions": [
        {
          "accessStrategies": [
            {
              "type": "open",
            },
          ],
          "mediaType": "application/json",
          "type": "asyncapi-v2",
          "url": "/.well-known/open-resource-discovery/v1/api-metadata/AdminService.asyncapi2.json",
        },
      ],
      "shortDescription": "Example ODM Event",
      "title": "ODM capjsord Events",
      "version": "1.0.0",
      "visibility": "public",
    },
    {
      "description": "This is an example event catalog that contains only a partial ODM cap-js-ord V1 event",
      "extensible": {
        "supported": "no",
      },
      "ordId": "capjs.ord:eventResource:undefined.CatalogService:v1",
      "partOfGroups": [
        "sap.cds:service:capjs.ord:undefined.CatalogService",
      ],
      "partOfPackage": "capjsord:package:undefined:v1",
      "releaseStatus": "beta",
      "resourceDefinitions": [
        {
          "accessStrategies": [
            {
              "type": "open",
            },
          ],
          "mediaType": "application/json",
          "type": "asyncapi-v2",
          "url": "/.well-known/open-resource-discovery/v1/api-metadata/CatalogService.asyncapi2.json",
        },
      ],
      "shortDescription": "Example ODM Event",
      "title": "ODM capjsord Events",
      "version": "1.0.0",
      "visibility": "public",
    },
  ],
  "groups": [
    {
      "groupId": "sap.cds:service:capjs.ord:undefined.AdminService",
      "groupTypeId": "sap.cds:service",
      "title": "Admin Service Title",
    },
    {
      "groupId": "sap.cds:service:capjs.ord:undefined.CatalogService",
      "groupTypeId": "sap.cds:service",
      "title": "Catalog Service Title",
    },
  ],
  "openResourceDiscovery": "1.9",
  "packages": [
    {
      "description": "Description for capjs ord",
      "ordId": "capjsord:package:undefined:v1",
      "partOfProducts": [
        "customer:product:capjs.ord:",
      ],
      "shortDescription": "Short description for capjs ord",
      "title": "capjs ord",
      "vendor": "customer:vendor:Customer:",
      "version": "1.0.0",
    },
  ],
  "policyLevel": "none",
  "products": [
    {
      "ordId": "customer:product:cap.js.ord:",
      "shortDescription": "Description for cap js ord",
      "title": "cap js ord",
      "vendor": "customer:vendor:customer:",
    },
  ],
}
```

@Fannon

lib/ord.js
```diff
@@ -156,7 +150,19 @@ const fGetAPIResources = (csn, global,packageIds) => {
  * @param {Object} csn object
  * @returns {Array<object>} The Event Resources array.
  */
-const fGetEventResources = (csn, global,packageIds) => global.aEvents.map((srv) => fCreateEventResourceTemplate(srv, csn.definitions[srv], global, packageIds)).filter((resource) => resource !== null && resource !== undefined);
+const fGetEventResources = (csn, global,packageIds) => {
```
zongqichen, line 153:
```suggestion
const fGetEventResources = (csn, global, packageIds) => {
```

lib/ord.js
```diff
@@ -113,14 +112,9 @@ const fGetGroups = (csn, global) => {
     // storing the group ids in a set to avoid duplicates
     let groupIds = new Set();
 
-    let serviceGroups = global.aServices
+    return global.aServices
```
zongqichen, line 115: I know it's not your problem, but this function implementation is really wired:
1. We should avoid using global to store data. 
2. If we have to use global, we should not pass it as argument, which against the usage of global. 
3. The name 'global.aServices' is really confused. I would suggest that, later we use oop to refactor it. 
4. The function name and return are also confused. The name is fGetGroups, and I would expect to get the list of group. But the return value is indirectly indication. I would suggest, either we define a variable: 'const groupList' and return this variable, or remove line "let groupIds = new Set();", e.g 
```javascript
const fGetGroups = (csn, global) => return global.aServices
 .map((serviceName) => fCreateGroupsTemplateForService(serviceName, csn.definitions[serviceName], new Set()))
 .filter((resource) => resource !== null && resource !== undefined);
```

lib/ord.js
```diff
@@ -156,7 +150,19 @@ const fGetAPIResources = (csn, global,packageIds) => {
...
+
+    const services = []
+    for (const srv of global.aServices) {
+        const hasEvents = global.aEvents.find((eventName) => eventName.startsWith(srv));
+        if (hasEvents) {
+            services.push(srv)
+        }
+    }
+
+    return services.map((srv) => fCreateEventResourceTemplate(srv, csn.definitions[srv], global, packageIds))
```
zongqichen, line 164:
```suggestion
    for (const serviceName of global.aServices) {
        const hasEvents = global.aEvents.find((eventName) => eventName.startsWith(serviceName));
        if (hasEvents) {
            services.push(serviceName)
        }
    }

    return services.map((serviceName) => fCreateEventResourceTemplate(serviceName, csn.definitions[serviceName], global, packageIds))
```
I think `srv` implies to serviceName. I would suggest use serviceNmae. I like `hasEvents`, is clear and good name. I found another problem, we don't have logger. In principle, for example line 154, if the event is empty, we return empty list, in line 164 we might also turn emtpy list if aServices is emtpy. To distinguish two conditions but same return value, we need logger to help us. Just an idea for the future.

__tests__/ord.test.js
```diff
@@ -13,4 +13,37 @@ describe("Tests for default ORD document", () => {
...
+  
+    test("The CDS Service Group ID includes the CDS Service identifier", () => {
+      for (const eventResource of document.eventResources) {
+        const [groupId] = eventResource.partOfGroups
+        expect(groupId).toMatch(GROUP_ID_REGEX)
+
+        const match = GROUP_ID_REGEX.exec(groupId)
+        if (match && match.groups?.service) {
+          let service = match.groups?.service
+          if (service.startsWith("undefined")) service = service.replace("undefined.", "")
```
zongqichen, line 41: Could you explain these two lines? And service seems to imply to serviceName?

ductaily: Yes, service imply serviceName, you can see it in the group id regex as a named capturing group.

The implementation here is not finished but my goal was to find the corresponding service in the CSN definition using the information I have in the ORD Document. In this case, I want to find that the eventResource `partOfGroups` is assigned to a `groupId` that is actually a service in the CSN.

Since they are attaching the `namespace` in the groupId, I want to remove the namespace part here to find the corresponding CSN definition. In this case, `undefined`.

I will improve this to check for the namespace and remove the namespace instead of `undefined`.

I hope this answered the question.

lib/ord.js
zongqichen: Thx @ductaily, I see https://github.com/cap-js/ord/blob/main/lib/ord.js#L163, the global is the name of object. This is what I want, but the name is conflict with javascript keyword. We should avoid it. And good point, the naming pattern for example **a**Services and **f**GetGroups, we should have agreement in discussion.

ductaily:
The tests are failing and I can't reproduce the error on @zongqichen and my machine.

@RoshniNaveenaS Do you have an idea what could be the cause and how we could resolve it?

ductaily:
> The tests are failing and I can't reproduce the error on @zongqichen and my machine.
> 
> @RoshniNaveenaS Do you have an idea what could be the cause and how we could resolve it?

I see now that the failure is caused by the new CDS Version: https://github.tools.sap/cap/cds/releases/tag/rel%2F8.2.0
The LinkedDefinitions were changed and do not contain `map`, `forEach`, ... anymore.

zongqichen:
> > The tests are failing and I can't reproduce the error on @zongqichen and my machine.
> > @RoshniNaveenaS Do you have an idea what could be the cause and how we could resolve it?
> 
> I see now that the failure is caused by the new CDS Version: https://github.tools.sap/cap/cds/releases/tag/rel%2F8.2.0 The LinkedDefinitions were changed and do not contain `map`, `forEach`, ... anymore.

Cool, your commit has fixed it. I have one more question, it seems github action will always install the latest version of cds, and brings the issue like it. We should control peerDepdencies upper limits as well?

zongqichen (approved):

lib/templates.js
```diff
@@ -289,7 +289,7 @@ const fCreateEventResourceTemplate = (srv, srvDefinition, global,packageIds) =>
  */
 function checkEntityFunctionAction(srvDefinition, global) {
   if (srvDefinition.entities) {
-    return srvDefinition.entities.map((entity) => {
+    return [...srvDefinition.entities].map((entity) => {
```
zongqichen, line 292: I think there are 5 more places need to be adjust, for example srvDefinition.actions.map. But we don't have corresponding test, lets ignore it in this pr.

__tests__/ord.test.js
```diff
@@ -13,4 +13,37 @@ describe("Tests for default ORD document", () => {
     const document = ord(csn);
     expect(document).toMatchSnapshot();
   });
+
+
+  describe("eventResources", () => {
+    const GROUP_ID_REGEX = /^([a-z0-9-]+(?:[.][a-z0-9-]+)*):([a-zA-Z0-9._\-/]+):([a-z0-9-]+(?:[.][a-z0-9-]+)*):(?<service>[a-zA-Z0-9._\-/]+)$/
```
Fannon, line 19: I like this idea to have named groups in a regexp!

lib/ord.js
Fannon: Just to document something that @zongqichen  and I discussed via call:

I don't think that `global` is a global variable. It's very likely a variable that the CAP Framework passes into the plugins, to pass a global context, where the CAP framework keeps the overall information on what is part of the current application. This is where you find a list of all services, entities etc.

ductaily:
> > > The tests are failing and I can't reproduce the error on @zongqichen and my machine.
> > > @RoshniNaveenaS Do you have an idea what could be the cause and how we could resolve it?
> > 
> > 
> > I see now that the failure is caused by the new CDS Version: https://github.tools.sap/cap/cds/releases/tag/rel%2F8.2.0 The LinkedDefinitions were changed and do not contain `map`, `forEach`, ... anymore.
> 
> Cool, your commit has fixed it. I have one more question, it seems github action will always install the latest version of cds, and brings the issue like it. We should control peerDepdencies upper limits as well?

@zongqichen Yes, we should specify the exact version.

__tests__/__snapshots__/ord.test.js.snap
```diff
@@ -100,6 +100,64 @@ exports[`Tests for default ORD document Successfully create ORD Documents with d
     },
   ],
   "description": "this is an application description",
+  "eventResources": [
+    {
+      "description": "This is an example event catalog that contains only a partial ODM cap-js-ord V1 event",
+      "extensible": {
+        "supported": "no",
+      },
+      "ordId": "capjs.ord:eventResource:undefined.AdminService:v1",
```
Fannon, line 109: I still feel bad about all those `undefined` here. But this is planned to be fixed with a follow up ADR?

zongqichen: Yeah, I will fix it in next PR

ductaily: The `undefined` issues are all related and coming from the `global.capNamespace`. That will also fix this.

aramovic79:
@ductaily: There are a few merge conflicts to be resolved...

@zongqichen @Fannon

ductaily:
> @ductaily: There are a few merge conflicts to be resolved...
> 
> @zongqichen @Fannon

Done.

aramovic79:
@RoshniNaveenaS @RamIndia : Can we merge this PR?
cc: @ductaily @zongqichen @Fannon

aramovic79:
@ductaily : I guess the following one is solved as well, correct?
`The groups for events shall be services, not the individual events → but then the last two entries in the bookstore sample are duplicate`

ductaily:
> @ductaily : I guess the following one is solved as well, correct? `The groups for events shall be services, not the individual events → but then the last two entries in the bookstore sample are duplicate`

Correct. This PR would also solve that issue.

PR #59, opened 2024-09-18
fix: Use appName as package name in the apiResources.partOfPackage

aramovic79:
Addressing the following issue given in the issue #38 

> "partOfPackage": "customer.capirebookstore:package:capire.bookshop:v1” → correct namespace prefix in package ref - use appName as packageName as a fallback.

### DIFF
--- a/__tests__/__snapshots__/ord.test.js.snap
+++ b/__tests__/__snapshots__/ord.test.js.snap
@@ -173,13 +173,13 @@ exports[`Tests for default ORD document Successfully create ORD Documents with d
   "openResourceDiscovery": "1.9",
   "packages": [
     {
-      "description": "Description for capjs ord",
+      "description": "Description for cap js ord",
       "ordId": "capjsord:package:undefined:v1",
       "partOfProducts": [
-        "customer:product:capjs.ord:",
+        "customer:product:cap.js.ord:",
       ],
-      "shortDescription": "Short description for capjs ord",
-      "title": "capjs ord",
+      "shortDescription": "Short description for cap js ord",
+      "title": "cap js ord",
       "vendor": "customer:vendor:Customer:",
       "version": "1.0.0",
     },
--- a/__tests__/__snapshots__/ord.test.js.snap
+++ b/__tests__/__snapshots__/ord.test.js.snap
@@ -18,11 +18,11 @@ exports[`Tests for default ORD document Successfully create ORD Documents with d
       "extensible": {
         "supported": "no",
       },
-      "ordId": "bookshop:apiResource:undefined.AdminService:v1",
+      "ordId": "capire.bookshopordsample:apiResource:undefined.AdminService:v1",
       "partOfGroups": [
-        "sap.cds:service:bookshop:undefined.AdminService",
+        "sap.cds:service:capire.bookshopordsample:undefined.AdminService",
       ],
-      "partOfPackage": "bookshop:package:undefined:v1",
+      "partOfPackage": "capirebookshopordsample:package:undefined:v1",
       "releaseStatus": "active",
       "resourceDefinitions": [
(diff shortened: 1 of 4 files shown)


RoshniNaveenaS (approved):

lib/ord.js
```diff
@@ -119,7 +119,7 @@ const fGetGroups = (csn, global) => {
  * Hierarchy to check data: cdsrc.json > defaults
  * @returns {Array<object>} The packages array.
  */
-const fGetPackages = (policyLevel,global) => global.env?.packages || (global.aEvents.length) ? defaults.packages(global.namespace,policyLevel,global.capNamespace) : defaults.packages(global.namespace,policyLevel,global.capNamespace).slice(0, 1)
+const fGetPackages = (policyLevel,global) => global.env?.packages || (global.aEvents.length) ? defaults.packages(global.appName,policyLevel,global.capNamespace) : defaults.packages(global.appName,policyLevel,global.capNamespace).slice(0, 1)
```
ductaily, line 122: We probably have to add tests to check if the package ORD IDs and partOfPackages in e.g., apiResources still match in different cases after this change.

__tests__/__snapshots__/ord.test.js.snap
```diff
@@ -173,13 +173,13 @@ exports[`Tests for default ORD document Successfully create ORD Documents with d
   "openResourceDiscovery": "1.9",
   "packages": [
     {
-      "description": "Description for capjs ord",
+      "description": "Description for cap js ord",
```
RamIndia, line 176: You are using ord sample project in xmpl project? 
@capire/ord-sample ?

ductaily: No, in the test directory we added a bookshop project which we load before running the test.

__tests__/ord.test.js
```diff
@@ -14,6 +14,27 @@ describe("Tests for default ORD document", () => {
...
+        expect(apiResource.partOfPackage).toMatch(PACKAGE_ID_REGEX)
+      }
+    })
+
+    test("The partOfPackage references an existing package", () => {
+        for (const apiResource of document.apiResources) {
+          expect(document.packages.find((pck) => pck.ordId == apiResource.partOfPackage)).toBeDefined
+        }
+      })
+  })
```
aramovic79, line 37: Two checks are added:

1. If the partOfPackage is a valid ORD ID
2. If it references an existing package.ordId
I guess there is no need to check if the packages[].ordId is a valid ORD ID because it is implicitly "covered" by the combination of these two checks.

@ductaily @Fannon @RamIndia : Do you agree or I should add some more tests/checks?

__tests__/ord.test.js
```diff
@@ -14,6 +14,27 @@ describe("Tests for default ORD document", () => {
...
+
+    test("partOfPackage values are valid ORD IDs ", () => {
+      for (const apiResource of document.apiResources) {
+        expect(apiResource.partOfPackage).toMatch(PACKAGE_ID_REGEX)
+      }
+    })
+
+    test("The partOfPackage references an existing package", () => {
+        for (const apiResource of document.apiResources) {
+          expect(document.packages.find((pck) => pck.ordId == apiResource.partOfPackage)).toBeDefined
```
ductaily, line 34:
```suggestion
          expect(document.packages.find((pck) => pck.ordId === apiResource.partOfPackage)).toBeDefined()
```

ductaily (approved):

aramovic79:
@RamIndia @RoshniNaveenaS : Please let us know if we may merge this PR. Thanks!

cc: @zongqichen @ductaily @Fannon

RamIndia (approved):
Looks Good

PR #60, opened 2024-09-18
fix: Use the fully-qualified names from .definitions instead of namespace

aramovic79:
Addressing the following issue from #38 :

> Never use CSN's .namespace property but always just use the fully-qualified names from .definitions

Instead of `csn.namespace`, we should use e.g. `cds.model.definitions["AdminService"].model.namespace`

### DIFF
--- a/__tests__/__snapshots__/ord.test.js.snap
+++ b/__tests__/__snapshots__/ord.test.js.snap
@@ -22,7 +22,7 @@ exports[`Tests for default ORD document Successfully create ORD Documents with d
       "partOfGroups": [
         "sap.cds:service:capire.bookshopordsample:AdminService",
       ],
-      "partOfPackage": "capirebookshopordsample:package:capire.bookshopordsample:v1",
+      "partOfPackage": "capire.bookshopordsample:package:customer.capirebookshopordsample:v1",
       "releaseStatus": "active",
       "resourceDefinitions": [
         {
@@ -69,7 +69,7 @@ exports[`Tests for default ORD document Successfully create ORD Documents with d
       "partOfGroups": [
         "sap.cds:service:capire.bookshopordsample:CatalogService",
       ],
-      "partOfPackage": "capirebookshopordsample:package:capire.bookshopordsample:v1",
+      "partOfPackage": "capire.bookshopordsample:package:customer.capirebookshopordsample:v1",
       "releaseStatus": "active",
       "resourceDefinitions": [
         {
@@ -110,7 +110,7 @@ exports[`Tests for default ORD document Successfully create ORD Documents with d
       "partOfGroups": [
         "sap.cds:service:capire.bookshopordsample:AdminService",
       ],
-      "partOfPackage": "capirebookshopordsample:package:capire.bookshopordsample:v1",
+      "partOfPackage": "capire.bookshopordsample:package:customer.capirebookshopordsample:v1",
       "releaseStatus": "beta",
       "resourceDefinitions": [
         {
@@ -138,7 +138,7 @@ exports[`Tests for default ORD document Successfully create ORD Documents with d
       "partOfGroups": [
         "sap.cds:service:capire.bookshopordsample:CatalogService",
       ],
-      "partOfPackage": "capirebookshopordsample:package:capire.bookshopordsample:v1",
+      "partOfPackage": "capire.bookshopordsample:package:customer.capirebookshopordsample:v1",
       "releaseStatus": "beta",
       "resourceDefinitions": [
         {
@@ -174,7 +174,7 @@ exports[`Tests for default ORD document Successfully create ORD Documents with d
   "packages": [
     {
       "description": "Description for capire bookshop ord sample",
-      "ordId": "capirebookshopordsample:package:capire.bookshopordsample:v1",
+      "ordId": "capire.bookshopordsample:package:customer.capirebookshopordsample:v1",
       "partOfProducts": [
         "customer:product:capire.bookshop.ord.sample:",
       ],
--- a/__tests__/__snapshots__/ord.test.js.snap
+++ b/__tests__/__snapshots__/ord.test.js.snap
@@ -18,9 +18,9 @@ exports[`Tests for default ORD document Successfully create ORD Documents with d
       "extensible": {
         "supported": "no",
       },
-      "ordId": "capire.bookshopordsample:apiResource:undefined.AdminService:v1",
+      "ordId": "capire.bookshopordsample:apiResource:AdminService:v1",
       "partOfGroups": [
-        "sap.cds:service:capire.bookshopordsample:undefined.AdminService",
+        "sap.cds:service:capire.bookshopordsample:AdminService",
       ],
       "partOfPackage": "capirebookshopordsample:package:undefined:v1",
       "releaseStatus": "active",
@@ -65,9 +65,9 @@ exports[`Tests for default ORD document Successfully create ORD Documents with d
       "extensible": {
         "supported": "no",
       },
-      "ordId": "capire.bookshopordsample:apiResource:undefined.CatalogService:v1",
+      "ordId": "capire.bookshopordsample:apiResource:CatalogService:v1",
       "partOfGroups": [
-        "sap.cds:service:capire.bookshopordsample:undefined.CatalogService",
+        "sap.cds:service:capire.bookshopordsample:CatalogService",
       ],
       "partOfPackage": "capirebookshopordsample:package:undefined:v1",
       "releaseStatus": "active",
@@ -106,9 +106,9 @@ exports[`Tests for default ORD document Successfully create ORD Documents with d
       "extensible": {
         "supported": "no",
       },
-      "ordId": "capire.bookshopordsample:eventResource:undefined.AdminService:v1",
+      "ordId": "capire.bookshopordsample:eventResource:AdminService:v1",
       "partOfGroups": [
-        "sap.cds:service:capire.bookshopordsample:undefined.AdminService",
+        "sap.cds:service:capire.bookshopordsample:AdminService",
       ],
       "partOfPackage": "capirebookshopordsample:package:undefined:v1",
       "releaseStatus": "beta",
@@ -134,9 +134,9 @@ exports[`Tests for default ORD document Successfully create ORD Documents with d
       "extensible": {
         "supported": "no",
       },
-      "ordId": "capire.bookshopordsample:eventResource:undefined.CatalogService:v1",
+      "ordId": "capire.bookshopordsample:eventResource:CatalogService:v1",
       "partOfGroups": [
-        "sap.cds:service:capire.bookshopordsample:undefined.CatalogService",
+        "sap.cds:service:capire.bookshopordsample:CatalogService",
       ],
       "partOfPackage": "capirebookshopordsample:package:undefined:v1",
       "releaseStatus": "beta",
@@ -160,12 +160,12 @@ exports[`Tests for default ORD document Successfully create ORD Documents with d
   ],
   "groups": [
     {
-      "groupId": "sap.cds:service:capire.bookshopordsample:undefined.AdminService",
+      "groupId": "sap.cds:service:capire.bookshopordsample:AdminService",
       "groupTypeId": "sap.cds:service",
       "title": "Admin Service",
     },
     {
-      "groupId": "sap.cds:service:capire.bookshopordsample:undefined.CatalogService",
+      "groupId": "sap.cds:service:capire.bookshopordsample:CatalogService",
       "groupTypeId": "sap.cds:service",
       "title": "Catalog Service",
     },
--- a/__tests__/__snapshots__/ord.test.js.snap
+++ b/__tests__/__snapshots__/ord.test.js.snap
@@ -22,7 +22,7 @@ exports[`Tests for default ORD document Successfully create ORD Documents with d
       "partOfGroups": [
         "sap.cds:service:capire.bookshopordsample:AdminService",
       ],
-      "partOfPackage": "capire.bookshopordsample:package:customer.capirebookshopordsample:v1",
+      "partOfPackage": "customer.capirebookshopordsample:package:capirebookshopordsample:v1",
       "releaseStatus": "active",
       "resourceDefinitions": [
         {
@@ -69,7 +69,7 @@ exports[`Tests for default ORD document Successfully create ORD Documents with d
       "partOfGroups": [
         "sap.cds:service:capire.bookshopordsample:CatalogService",
       ],
-      "partOfPackage": "capire.bookshopordsample:package:customer.capirebookshopordsample:v1",
+      "partOfPackage": "customer.capirebookshopordsample:package:capirebookshopordsample:v1",
       "releaseStatus": "active",
       "resourceDefinitions": [
         {
@@ -110,7 +110,7 @@ exports[`Tests for default ORD document Successfully create ORD Documents with d
       "partOfGroups": [
         "sap.cds:service:capire.bookshopordsample:AdminService",
       ],
-      "partOfPackage": "capire.bookshopordsample:package:customer.capirebookshopordsample:v1",
+      "partOfPackage": "customer.capirebookshopordsample:package:capirebookshopordsample:v1",
       "releaseStatus": "beta",
       "resourceDefinitions": [
         {
@@ -138,7 +138,7 @@ exports[`Tests for default ORD document Successfully create ORD Documents with d
       "partOfGroups": [
         "sap.cds:service:capire.bookshopordsample:CatalogService",
       ],
-      "partOfPackage": "capire.bookshopordsample:package:customer.capirebookshopordsample:v1",
+      "partOfPackage": "customer.capirebookshopordsample:package:capirebookshopordsample:v1",
       "releaseStatus": "beta",
       "resourceDefinitions": [
         {
@@ -174,7 +174,7 @@ exports[`Tests for default ORD document Successfully create ORD Documents with d
   "packages": [
     {
       "description": "Description for capire bookshop ord sample",
-      "ordId": "capire.bookshopordsample:package:customer.capirebookshopordsample:v1",
+      "ordId": "customer.capirebookshopordsample:package:capirebookshopordsample:v1",
       "partOfProducts": [
         "customer:product:capire.bookshop.ord.sample:",
       ],
--- a/__tests__/__snapshots__/ord.test.js.snap
+++ b/__tests__/__snapshots__/ord.test.js.snap
@@ -22,7 +22,7 @@ exports[`Tests for default ORD document Successfully create ORD Documents with d
       "partOfGroups": [
         "sap.cds:service:capire.bookshopordsample:AdminService",
       ],
-      "partOfPackage": "capirebookshopordsample:package:undefined:v1",
+      "partOfPackage": "capirebookshopordsample:package:capire.bookshopordsample:v1",
       "releaseStatus": "active",
       "resourceDefinitions": [
         {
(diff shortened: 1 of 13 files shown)


lib/templates.js
```diff
@@ -118,18 +118,13 @@ function _getTitleFromServiceName(srv) {
 const fCreateGroupsTemplateForService = (srv, srvDefinition) => {
     const ordExtensions = fReadORDExtensions(srvDefinition);
 
-  let fullyQualifiedServiceName = srv;
-  if (!srv.includes(global.capNamespace)) {
-    fullyQualifiedServiceName = global.capNamespace + "." + srv;
-  }
```
aramovic79, line 124: No need for generating fullyQualifiedServiceName: just take the name of the service as it is generated by the cds compiler.

cc: @Fannon @zongqichen

lib/defaults.js
```diff
@@ -33,10 +33,10 @@ module.exports = {
     },
   ],
   groupTypeId: "sap.cds:service",
-  packages: function getPackageData (name, policyLevel,capNamespace) {
+  packages: function getPackageData (name, policyLevel,namespace) {
```
aramovic79, line 36: When generating the name of package, we are now using `namespace`(taken from .cdsrc.json and package.json/name as fallback) instead of `capNamespace`(which is `csn.namespace` and can be undefined in case when the namespace is not given in service definition).

@Fannon @zongqichen : Do you agree?

lib/ord.js
```diff
@@ -29,8 +29,7 @@ const fInitializeGlobal = (csn) => {
     const aServices = [];
     const aODMEntity = [];
 
-    const capNamespace = csn.namespace;
-     // namespace variable value if present in cdsrc.json take it there or else from package.json
```
aramovic79, line 33: Not needed if we use `global.namespace`(value given in .cdsrc.json || package.json). 

@Fannon @zongqichen : Do you agree?

lib/ord.js
```diff
@@ -43,7 +42,6 @@ const fInitializeGlobal = (csn) => {
     const fServicesFilter = (keyDefinition) => keyDefinition.kind === "service" && !keyDefinition['@cds.external'];
     const fODMEntityFilter = (key, keyDefinition) => {
         return keyDefinition.kind === "entity"
-            && key.includes(capNamespace)
```
aramovic79, line 46: This wasn't correct in case when the `namespace` is not defined in the service definition file. In that case, the `capNamespace` will be **undefined** and this filter will be wrong.
Another option would be: `&& (capNamespace && key.includes(capNamespace)` instead of this line.
@Fannon @zongqichen @RamIndia : Which option do you prefer(if any)?

lib/ord.js
```diff
@@ -116,10 +113,10 @@ const fGetGroups = (csn, global) => {
 
 /**
  * Retrieves the packages.
- * Hierarchy to check data: cdsrc.json > defaults
+ * Hierarchy to check data: cdsrc.json > defaults.
  * @returns {Array<object>} The packages array.
  */
-const fGetPackages = (policyLevel,global) => global.env?.packages || (global.aEvents.length) ? defaults.packages(global.appName,policyLevel,global.capNamespace) : defaults.packages(global.appName,policyLevel,global.capNamespace).slice(0, 1)
+const fGetPackages = (policyLevel,global) => global.env?.packages || (global.aEvents.length) ? defaults.packages(global.appName,policyLevel,global.namespace) : defaults.packages(global.appName,policyLevel,global.namespace).slice(0, 1)
```
aramovic79, line 119: Similar as in case of groups, we use `namespace` instead of **cap**Namespace.

aramovic79:
@RoshniNaveenaS @Fannon @zongqichen @RamIndia : Could you please review this pull request? Thanks.

lib/ord.js
zongqichen: Hi @aramovic79 , do you know why there are two namespaces and what is the context behind them?

aramovic79: > Hi @aramovic79 , do you know why there are two namespaces and what is the context behind them?

AFAIK, the `capNamespace` is a namespaces as given in the service.cds definition file. If it is not given(which can be the case), this value is undefined, of course; on the other side, `namespace` is the value given in the .cdsrc.json file and as a fallback, we are using name of the package from the package.json file. 
cc: @RoshniNaveenaS @RamIndia @Fannon

__tests__/__snapshots__/ord.test.js.snap
```diff
@@ -174,7 +174,7 @@ exports[`Tests for default ORD document Successfully create ORD Documents with d
   "packages": [
     {
       "description": "Description for capire bookshop ord sample",
-      "ordId": "capirebookshopordsample:package:undefined:v1",
+      "ordId": "capirebookshopordsample:package:capire.bookshopordsample:v1",
```
zongqichen, line 177: Hi @Fannon, is this package id valid? `<namespace>:<conceptName>:<resourceName>:[v<majorVersion>]`, namesapce and resourceName are reverted?

lib/defaults.js
```diff
@@ -33,10 +33,10 @@ module.exports = {
...
     function createPackage(name, tag) {
       return {
-        ordId: `${regexWithRemoval(name)}:package:${capNamespace}${tag}`,
+        ordId: `${regexWithRemoval(name)}:package:${namespace}${tag}`,
```
zongqichen, line 39: Is this package ordId valid? `<namespace>:<conceptName>:<resourceName>:[v<majorVersion>]`, the last part is `${namespace}${tag}`?

I also suggest provides variables as context, and extract it into function, for example:
```suggestion
        ordId: getPackageOrdID(),
       ...
       function getPackageOrdID() {
         const namespace = namesapce;
         const conceptName = package;
         const resourceName = sourceName;
         const version = version:
         return `${namespace}:${conceptName}:${resourceName}:${version}:`
       }
```

__tests__/__snapshots__/ord.test.js.snap
Fannon: No, this is not valid. The first fragment is not an ORD namespace. It should have the same namespace as you define in the cdsrc file. 

The third fragment (resource name) looks fine. This is just a fully qualified name from CAP view, not necessarily a namespace.

lib/defaults.js
Fannon: See above. You need to use the globally defined ORD namespace , not `regexWithRemoval(name)`

lib/ord.js
zongqichen: Thanks, I'm confused when namespace are different in definition file, .cdsrc.json and package.json. I also need other opinion.

lib/defaults.js
Fannon: Btw. you have typo with "namesapce"

lib/templates.js
```diff
@@ -223,7 +207,7 @@ const fCreateAPIResourceTemplate = (srv, srvDefinition, global, packageIds) => {
       let obj = {
         ordId: `${fReplaceSpecialCharacters(
           global.namespace
-        )}:apiResource:${srvName}:v1`,
```
zongqichen, line 226: I would suggest that extract into getApiResourceOrdId function and follow this rule: `<ordId> := <namespace>:<conceptName>:<resourceName>:[v<majorVersion>]`

lib/templates.js
```diff
@@ -118,18 +118,13 @@ function _getTitleFromServiceName(srv) {
...
-  }
-
     if(!srvDefinition) {
       console.warn("Unable to find service definition:", srv)
       return undefined
     }
 
     if (srvDefinition && checkEntityFunctionAction(srvDefinition, global).length > 0) {
-      let groupId = _getGroupID(fullyQualifiedServiceName, defaults.groupTypeId);
+      let groupId = _getGroupID(srv, defaults.groupTypeId);
```
zongqichen, line 127: What does `srv` mean here? Is the lastTwoFragments? I would suggest name it more context

lib/templates.js
```diff
@@ -158,14 +153,8 @@ function _getEventTitle(event) {
...
   const ordExtensions = fReadORDExtensions(eventDefinition);
-
-  let fullyQualifiedEventName = event;
-  if (!event.includes(global.capNamespace)) {
-    fullyQualifiedEventName = global.capNamespace + "." + event;
-  }
-
   let groupId = _getGroupID(
-    fullyQualifiedEventName,
+    event,
```
zongqichen, line 157: Same here, event is too general. I notice it comes from previous code. I would like give it meaningful name as well.

lib/templates.js
aramovic79: srv = service name; I've renamed it to the srvName? 
@zongqichen @Fannon @RamIndia : Sounds fine to you?

__tests__/__snapshots__/ord.test.js.snap
```diff
@@ -18,11 +18,11 @@ exports[`Tests for default ORD document Successfully create ORD Documents with d
...
         "supported": "no",
       },
-      "ordId": "capire.bookshopordsample:apiResource:undefined.AdminService:v1",
+      "ordId": "capire.bookshopordsample:apiResource:AdminService:v1",
       "partOfGroups": [
-        "sap.cds:service:capire.bookshopordsample:undefined.AdminService",
+        "sap.cds:service:capire.bookshopordsample:AdminService",
       ],
-      "partOfPackage": "capirebookshopordsample:package:undefined:v1",
+      "partOfPackage": "capirebookshopordsample:package:capire.bookshopordsample:v1",
```
Fannon, line 25: This is still wrong, it should be: 

```suggestion
      "partOfPackage": "capire.bookshopordsample:package:capirebookshopordsamplev1",
```

(not sure on the package "name" (third fragment). It could be concatnated a little bit more nicely, but that's cosmetic)

__tests__/__snapshots__/ord.test.js.snap
Fannon:
```suggestion
      "ordId": "capire.bookshopordsample:package:capirebookshopordsample:v1",
```

This is more likely the correct ORD ID.

__tests__/unittest/__snapshots__/templates.test.js.snap
```diff
@@ -42,9 +42,9 @@ exports[`templates fCreateAPIResourceTemplate should create API resource templat
     "extensible": {
       "supported": "no",
     },
-    "ordId": "customer:apiResource:undefined.MyService:v1",
+    "ordId": "customer:apiResource:MyService:v1",
```
Fannon, line 45: Why have we suddenly `customer` as a namespace here? This is just a vendor namespace, there needs to be a full system namespace at least, so `customer.something`. Here you would use the namespace from CDS RC

Fannon: Is this because in this test, we have no cdsrc and use full defaults? In that case we should default to something like `customer.<packageNameEscaped>`

lib/defaults.js
Fannon: Maybe we should name things more clearly. Like this is the ORD namespace (not a CSN namespace)

```suggestion
  packages: function getPackageData (name, policyLevel, ordNamespace) {
```

lib/templates.js
aramovic79: `srv` is now `serviceName`, `srvDefinition` is `serviceDefinition` and `event` is `eventName`.

__tests__/__snapshots__/ord.test.js.snap
aramovic79: As mentioned in the today's meeting, the package ordId is now: 
`"partOfPackage": "capire.bookshopordsample:package:customer.capirebookshopordsample:v1"`
So, @Fannon comparing to what you wrote in the last comment, the **customer.** prefix is added and **:** in front of the tag v1.
@Fannon : Looks ok now? @zongqichen ?

__tests__/unittest/__snapshots__/templates.test.js.snap
aramovic79: @Fannon @zongqichen :Renamed into `customer.testNamespace` since it was used in one of the assertions from above. Actually, the string `customer.testNamespace-123` was used, but it doesn't change the thing.

aramovic79:
@zongqichen @Fannon : Could you please review once more? Thanks.

__tests__/__snapshots__/ord.test.js.snap
```diff
@@ -65,11 +65,11 @@ exports[`Tests for default ORD document Successfully create ORD Documents with d
...
         "supported": "no",
       },
-      "ordId": "capire.bookshopordsample:apiResource:undefined.CatalogService:v1",
+      "ordId": "capire.bookshopordsample:apiResource:CatalogService:v1",
       "partOfGroups": [
-        "sap.cds:service:capire.bookshopordsample:undefined.CatalogService",
+        "sap.cds:service:capire.bookshopordsample:CatalogService",
       ],
-      "partOfPackage": "capirebookshopordsample:package:undefined:v1",
+      "partOfPackage": "capire.bookshopordsample:package:customer.capirebookshopordsample:v1",
```
Fannon, line 72: Sorry, I'm still confused. Is this the namespace as it was provided in the cdsrc file? I think it would be better to go with a `customer.bookshopsample` as it's more realistic.

aramovic79: This is the namespace as it is given in the package.json because in the bookshop test folder, we don't have .cdsrc.json file.
As you know, we are constructing namespace this way:
`cds.env["ord"]?.namespace || packageName.replace(/[^\w/]/g, "").replace(/\//, ".");`

Again, there is no .cdsrc.json file in the ORD/__tests__/bookshop folder, so the first part of OR statement is discarded;

Fannon: ok, but if we use the package name, we need to prepent it with `customer.` (as you started doing when we talked about it)

Fannon:
```
cds.env["ord"]?.namespace || 'customer.` + packageName.replace(/[^\w/]/g, "").replace(/\//, "");
```

aramovic79: This is how the packages looks like now in our test: 
```
  packages: [
    {
      ordId: "customer.capire.bookshopordsample:package:capirebookshopordsample:v1",
      title: "capire bookshop ord sample",
      shortDescription: "Short description for capire bookshop ord sample",
      description: "Description for capire bookshop ord sample",
      version: "1.0.0",
      partOfProducts: [
        "customer:product:capire.bookshop.ord.sample:",
      ],
      vendor: "customer:vendor:Customer:",
    },
  ],
```
@zongqichen  @Fannon : Looks correct now?

Fannon: no, there should not be `.capire` in the namespace, we also need to escape `.` away unfortunately. Correct would be: `customer.capirebookshopordsample`

Fannon: When we take the package name and make a default namespace out of it, we basically need to replace any non ASCII char with "". The result will be really ugly, that's why I hope that they choose their own namespaces explicitly

zongqichen: Hi @Fannon, why should we use the value from package.json? If there are muli packages generated, each package OrdId will be same? Could you explain more about this [sentences](https://pages.github.tools.sap/CentralEngineering/open-resource-discovery-specification/spec-v1#ord-id):
```
<resourceName> := the technical resource name.
MUST only contain ASCII letters (a-z, A-Z), digits (0-9) and the special characters -, _ and ..
```
1. It shows special characters `-` `_` `.` are allowed, but we use regex to filter out?
2. What's the meaning of technical resource name?

__tests__/__snapshots__/ord.test.js.snap
Fannon: > So, @Fannon comparing to what you wrote in the last comment, the customer. prefix is added and : in front of the tag v1.

Yes, `customer.` prefix ist still missing. If we have no explicit namepace set, it hsould be `customer.capirebookshopordsample:package:capirebookshopordsamplev1`

__tests__/__snapshots__/ord.test.js.snap
Fannon: > It shows special characters - _ . are allowed, but we use regex to filter out?

there are different rules for the namespaces than for the resource name. They each need different escaping rules. For example `.` is allowed in a Resource name, but cannot be allowed in a namepace because a `.` separates namespaces fragments from each other.

Fannon: Looks good! :)

lib/templates.js
```diff
@@ -265,14 +249,10 @@ const fCreateAPIResourceTemplate = (srv, srvDefinition, global, packageIds) => {
...
-  let srvName = srv;
-  if (!srvName.includes(global.capNamespace)) {
-    srvName = global.capNamespace + "." + srv;
-  }
   return {
     ordId: `${fReplaceSpecialCharacters(
-      global.namespace
-    )}:eventResource:${srvName}:v1`,
+      global.ordNamespace
+    )}:eventResource:${srv}:v1`,
```
zongqichen, line 255: provide meaning name of `srv`

lib/templates.js
```diff
@@ -284,8 +264,8 @@ const fCreateEventResourceTemplate = (srv, srvDefinition, global,packageIds) =>
                   "CAP Event resource describing events / messages.",
     version:  ordExtensions.version ?? "1.0.0",
     releaseStatus:  ordExtensions.releaseStatus ?? "beta",
-    partOfPackage: _getPackageID(global.capNamespace,packageIds,'event'),
-    partOfGroups: [_getGroupID(srvName, defaults.groupTypeId, global, false)],
+    partOfPackage: _getPackageID('customer.' + global.ordNamespace.replace(/[^a-zA-Z0-9]/g, ""),packageIds,'event'),
+    partOfGroups: [_getGroupID(srv, defaults.groupTypeId, global, false)],
```
zongqichen, line 268: same here, `srv`

lib/templates.js
```diff
@@ -118,18 +118,13 @@ function _getTitleFromServiceName(srv) {
...
-    fullyQualifiedServiceName = global.capNamespace + "." + srv;
-  }
-
     if(!srvDefinition) {
       console.warn("Unable to find service definition:", srv)
       return undefined
     }
 
     if (srvDefinition && checkEntityFunctionAction(srvDefinition, global).length > 0) {
-      let groupId = _getGroupID(fullyQualifiedServiceName, defaults.groupTypeId);
```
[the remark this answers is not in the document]
zongqichen: Move it to following refactor item

aramovic79:
@Fannon @zongqichen : Could you please review?

__tests__/__snapshots__/ordCdsrc.test.js.snap
```diff
@@ -0,0 +1,208 @@
...
+      "groupId": "sap.cds:service:sap.test.cdsrc.sample:CatalogService",
+      "groupTypeId": "sap.cds:service",
+      "title": "Catalog Service",
+    },
+  ],
+  "openResourceDiscovery": "1.9",
+  "packages": [
+    {
+      "description": "Description for capire bookshop ord sample",
+      "ordId": "sap.test.cdsrc.sample:package:capirebookshopordsample-api:v1",
```
zongqichen, line 177: it looks it could generate multi packages by appending different suffix, I guess it's valid @Fannon.

zongqichen (approved):
LGTM

__tests__/__snapshots__/ordCdsrc.test.js.snap
Fannon: Ok, so does it now append suffixes or not? I thought for now, it just generates one package and assigns all the ORD resources to it? But in the future, we would add the option to automatically add the suffixes and assign the resources to the right kind of package :)

zongqichen: Yes, the [original code](https://github.com/cap-js/ord/blob/main/lib/templates.js#L349) has the logic to append suffix to different type.

Issue #49, opened 2024-08-28
Allow generic override / expand of ORD information on global level


Fannon:
Right now, the documentation in https://github.com/cap-js/ord/blob/main/ord.md indicates two ways to add ORD information: 

* Add global ord config, that is then applied to several places
* Override ORD information via Annotations on CDS Models like a Service. 

This still misses some flexibility:
* It's not possible to add new ORD resources or resource types that are not supported / provided by the CAP Framework right now (e.g. Data Products or Entity Types)
* It's not possible to override something in case this is necessary

### Suggestion

Add a completely generic way to provide partial ORD documents, e.g. like:

```js
// cdsrc.json or package.json cds context
  "ord": {
    "namespace": "sap.sample",
    "description": "this is my custom description",
    "policyLevel": "sap:core:v1",
    "customOrd": "./ord/custom.ord.json" // Link to static ORD JSON file to merge in
  }
```

Expected behavior: 
* The interface of the `custom` object is the same as the ORD Document interface. 
* The collections (like "dataProducts" are iterated and for those resource types that CAP supports we check if they are already defined through CAP. If yes, the custom definition wins (override)"
* All content in the `custom` object are merged to the final ORD Document that the CAP plugin produces.

zongqichen:
Hi @Fannon, it's a good proposal. I would suggest that renames *customOrd* -> *customOrdContentFile*, with more context.

Fannon:
> Hi @Fannon, it's a good proposal. I would suggest that renames _customOrd_ -> _customOrdContentFile_, with more context.

Sounds good!

zongqichen:
Use cases:
1. We want to add new ord resources that are not supported by cap framework, e.g. dataProducts
2. We need to enhance the list of generated ord resources, e.g. adding new pacakge or new apiResource
3. We need to patch the existing generated ord resources, e.g. customized the existing package
- override property(string, array, object) generated by the plugin
- add new property(string, array, object)
- delete property when it set to null (still be open how to cover it properly)

4. Not in scope:
- delete pre-genereated ord top level resources: e.g. packages, apiResources

Fannon:
We should completely follow the way how the upcoming ORD Diff documents work. See  https://github.tools.sap/CentralEngineering/open-resource-discovery-specification/pull/862/files

```md
ORD Delta Documents
ORD Delta Documents are an optional feature that allows to efficiently describe a "patch" / "diff" to the static baseline state.

[...]

PR #63, opened 2024-09-26
Feat: extend custom ORD content

zongqichen:


swennemers:
I was wondering, if we should have the custom ORD content be in sync with the ideas of a delta ORD document...
/open-resource-discovery-specification/pull/862/
@Fannon , what do you think?

Fannon:
> I was wondering, if we should have the custom ORD content be in sync with the ideas of a delta ORD document... /open-resource-discovery-specification/pull/862/ @Fannon , what do you think?

Do you mean that we should not just have a way to add / completely override, but use the same idea of the "JSON Patch" format (with explicit `null` semantics) from https://github.tools.sap/CentralEngineering/open-resource-discovery-specification/pull/862?

I'm not entirely sure. Merging sounds better first, but it would make it really difficult to completely override something that the CAP framework generated by default. If you just want to selectively override something, you would likely use the local annotations where they belong (Service level). 
We would also have to explain the merge behavior, allthough it's not that complicated (JSON Merge Patch essentially). 

So we need to pick between:

#### Merge Behavior
* If custom ORD content doesn't exist yet, just add it
* If custom ORD content has same ORD ID as existing content, merge it with the CAP generated content. The values in the override take precedence. If a value in override is set to `null` the property is to be removed entirely.

### Override Behavior
* If custom ORD content doesn't exist yet, just add it
* If custom ORD content has same ORD ID as existing content, replace the generated content entirely with the override.

swennemers:
I am not on the latest stage with the exact delta ORD syntax. 

What we observe here is that we need deep merge logic (e.g. overwrite an APIs package assignment without the need for rewriting everything). Don't we need as similar logic also for delta ORD, like overwriting just the endpoint of an API? 

I agree, that we need to explain the merge & overwrite logic. However, I thought, if we can sync these, we only need to describe it once

Fannon:
> I am not on the latest stage with the exact delta ORD syntax.
> 
> What we observe here is that we need deep merge logic (e.g. overwrite an APIs package assignment without the need for rewriting everything). Don't we need as similar logic also for delta ORD, like overwriting just the endpoint of an API?
> 
> I agree, that we need to explain the merge & overwrite logic. However, I thought, if we can sync these, we only need to describe it once

Ok, but overwriting a package assignment should not be done on the global ORD override level. You should do it on the CDS Service Level with `@ORD.Extensions`. Here we already have a merge logic, allthough not the best one (See issue #49)

swennemers:
> > I am not on the latest stage with the exact delta ORD syntax.
> > What we observe here is that we need deep merge logic (e.g. overwrite an APIs package assignment without the need for rewriting everything). Don't we need as similar logic also for delta ORD, like overwriting just the endpoint of an API?
> > I agree, that we need to explain the merge & overwrite logic. However, I thought, if we can sync these, we only need to describe it once
> 
> Ok, but overwriting a package assignment should not be done on the global ORD override level. You should do it on the CDS Service Level with `@ORD.Extensions`. Here we already have a merge logic, although not the best one (See issue #49)

true, bad example with package assignment. but not every ORD Resource has the CDS entities, e.g. package. what should we do there.

lib/extendOrdWithCustom.js
```diff
@@ -0,0 +1,60 @@
...
+    const existingContent = newOrdContent[key];
+    for (const subCustomObj of customORDContent[key]) {
+        const customContentSubObjOrdId = subCustomObj.ordId;
+        for (const existingContentSubObj of structuredClone(existingContent)) {
+            const existingContentSubObjOrdId = existingContentSubObj.ordId;
+            if (customContentSubObjOrdId === existingContentSubObjOrdId) {
+                throw new Error(`Detect conflict OrdId in custom ORD: ${key}, OrdId: ${customContentSubObjOrdId}, please resolve it.`);
+            } else {
+                existingContent.push(subCustomObj);
+                console.log(`Add custom ORD content: ${key}, OrdId: ${customContentSubObjOrdId}, newOrdContent: ${JSON.stringify(newOrdContent)}`);
```
swennemers, line 22:
```suggestion
                console.log(`Add custom ORD content: ${key}, ordId: ${customContentSubObjOrdId}, newOrdContent: ${JSON.stringify(newOrdContent)}`);
```

lib/extendOrdWithCustom.js
```diff
@@ -0,0 +1,60 @@
...
+}
+
+function handleExistingContentIfIsObject(newOrdContent, customORDContent, key) {
+    const existingContent = newOrdContent[key];
+    for (const subCustomObj of customORDContent[key]) {
+        const customContentSubObjOrdId = subCustomObj.ordId;
+        for (const existingContentSubObj of structuredClone(existingContent)) {
+            const existingContentSubObjOrdId = existingContentSubObj.ordId;
+            if (customContentSubObjOrdId === existingContentSubObjOrdId) {
+                throw new Error(`Detect conflict OrdId in custom ORD: ${key}, OrdId: ${customContentSubObjOrdId}, please resolve it.`);
```
swennemers, line 19:
```suggestion
                throw new Error(`Detect conflict ordId in custom ORD: ${key}, ordId: ${customContentSubObjOrdId}, please resolve it.`);
```

swennemers:
Let's handle the custom package assignment in https://github.com/cap-js/ord/issues/64

zongqichen:
> > I was wondering, if we should have the custom ORD content be in sync with the ideas of a delta ORD document... /open-resource-discovery-specification/pull/862/ @Fannon , what do you think?
> 
> Do you mean that we should not just have a way to add / completely override, but use the same idea of the "JSON Patch" format (with explicit `null` semantics) from https://github.tools.sap/CentralEngineering/open-resource-discovery-specification/pull/862?
> 
> I'm not entirely sure. Merging sounds better first, but it would make it really difficult to completely override something that the CAP framework generated by default. If you just want to selectively override something, you would likely use the local annotations where they belong (Service level). We would also have to explain the merge behavior, allthough it's not that complicated (JSON Merge Patch essentially).
> 
> So we need to pick between:
> 
> #### Merge Behavior
> * If custom ORD content doesn't exist yet, just add it
> * If custom ORD content has same ORD ID as existing content, merge it with the CAP generated content. The values in the override take precedence. If a value in override is set to `null` the property is to be removed entirely.
> 
> #### Override Behavior
> * If custom ORD content doesn't exist yet, just add it
> * If custom ORD content has same ORD ID as existing content, replace the generated content entirely with the override.

At the moment, I use lodash.merge as merge strategy: https://lodash.com/docs/4.17.15#merge. You can see the [code](https://github.com/cap-js/ord/blob/add_override_ord_information/lib/extendOrdWithCustom.js#L24). It will override existing item and merge new item if not existed.

__tests__/bookshop/ord/custom.ord.json
```diff
@@ -0,0 +1,51 @@
...
+            "description": "new AdminService",
+            "ordId": "sap.test.cdsrc.sample:apiResource:AdminService:v1",
+            "entityTypeMappings": [
+                {
+                    "entityTypeTargets": [
+                        "new.entityType.target"
+                    ]
+                }
+            ]
+        }
```
swennemers, line 23: Better have a testcase, where you add a new rest API, the other API resource overrides should be handled via @ORD.Extension mechanism, see linked issue.

Fannon:
> At the moment, I use lodash.merge as merge strategy: https://lodash.com/docs/4.17.15#merge. You can see the [code](https://github.com/cap-js/ord/blob/add_override_ord_information/lib/extendOrdWithCustom.js#L24). It will override existing item and merge new item if not existed.

By default the lodash merge will not have the JSON merge patch semantics with `null`. You need to make sure to also cover that (please also add tests for this, as it's easy to get this wrong). 

Here is some JSFiddle that I wrote for showing how AsyncAPI trait inheritance could work: https://jsfiddle.net/FannonF/yLsonr57/latest/

zongqichen:
> > At the moment, I use lodash.merge as merge strategy: https://lodash.com/docs/4.17.15#merge. You can see the [code](https://github.com/cap-js/ord/blob/add_override_ord_information/lib/extendOrdWithCustom.js#L24). It will override existing item and merge new item if not existed.
> 
> By default the lodash merge will not have the JSON merge patch semantics with `null`. You need to make sure to also cover that (please also add tests for this, as it's easy to get this wrong).
> 
> Here is some JSFiddle that I wrote for showing how AsyncAPI trait inheritance could work: https://jsfiddle.net/FannonF/yLsonr57/latest/
> 
> ```js
> function apply(origin, patch) {
>   if (!isObject(patch)) {
>     // If the patch is not an object, it replaces the origin.
>     return patch;
>   }
> 
>   const result = !isObject(origin) ? // Non objects are being replaced.
>     {} : // Make sure we never modify the origin.
>     Object.assign({}, origin);
> 
>   Object.keys(patch).forEach(key => {
>     const patchVal = patch[key];
>     if (patchVal === null) {
>       delete result[key];
>     } else {
>       result[key] = apply(result[key], patchVal);
>     }
>   });
>   return result;
> }
> ```
Thanks, I forget this requirement.

zongqichen:
> > At the moment, I use lodash.merge as merge strategy: https://lodash.com/docs/4.17.15#merge. You can see the [code](https://github.com/cap-js/ord/blob/add_override_ord_information/lib/extendOrdWithCustom.js#L24). It will override existing item and merge new item if not existed.
> 
> By default the lodash merge will not have the JSON merge patch semantics with `null`. You need to make sure to also cover that (please also add tests for this, as it's easy to get this wrong).
> 
> Here is some JSFiddle that I wrote for showing how AsyncAPI trait inheritance could work: https://jsfiddle.net/FannonF/yLsonr57/latest/
> 
> ```js
> function apply(origin, patch) {
>   if (!isObject(patch)) {
>     // If the patch is not an object, it replaces the origin.
>     return patch;
>   }
> 
>   const result = !isObject(origin) ? // Non objects are being replaced.
>     {} : // Make sure we never modify the origin.
>     Object.assign({}, origin);
> 
>   Object.keys(patch).forEach(key => {
>     const patchVal = patch[key];
>     if (patchVal === null) {
>       delete result[key];
>     } else {
>       result[key] = apply(result[key], patchVal);
>     }
>   });
>   return result;
> }
> ```

Hi @Fannon, I have some corner cases need to discuss:
1. if set null in array, should I remove the whole array property? e.g:
```json
{
    "apiResources": [
        {
            "ordId": "sap.sm:apiResource:SupplierService:v1",
            "partOfGroups": [
                null,
                "sap.cds:service:sap.test.cdsrc.sample:AdminService"
            ],
            "partOfPackage": "sap.sm:package:smDataProducts:v1"
        }
    ]
}
```
I should remove the ´partOfGroups´?

2. if null set in null in default property array, should I leave it as empty or removed?
```json
{
    "apiResources": [
        {null},
        {
            "ordId": "sap.sm:apiResource:SupplierService:v1",
            "partOfGroups": [
                "sap.cds:service:sap.test.cdsrc.sample:AdminService"
            ],
            "partOfPackage": "sap.sm:package:smDataProducts:v1"
        }
    ]
}
```
or this case:
```json
{
    "apiResources": [
         null
    ]
}
```

Fannon:
Nice summary of corner-cases, @zongqichen . I'll answer inline below:

> Hi @Fannon, I have some corner cases need to discuss:
> 
> 1. if set `null` in array, should I remove the whole array property? e.g:
> 
> ```json
> {
>     "apiResources": [
>         {
>             "ordId": "sap.sm:apiResource:SupplierService:v1",
>             "partOfGroups": [
>                 null,
>                 "sap.cds:service:sap.test.cdsrc.sample:AdminService"
>             ],
>             "partOfPackage": "sap.sm:package:smDataProducts:v1"
>         }
>     ]
> }
> ```

To keep it simple (and stick with JSON Merge patch), we should always override arrays completely. Everything else will get complicated very quickly when ordering changes. Setting `null` inside an array is therefore invalid, as it would result in having a `null` entry in an array, which is not allowed in ORD.

> 
> I should remove the ´partOfGroups´?
> 
> 2. if `null `set in default property array, should I leave it as empty or removed?
> 
> ```json
> {
>     "apiResources": [
>         {null},
>         {
>             "ordId": "sap.sm:apiResource:SupplierService:v1",
>             "partOfGroups": [
>                 "sap.cds:service:sap.test.cdsrc.sample:AdminService"
>             ],
>             "partOfPackage": "sap.sm:package:smDataProducts:v1"
>         }
>     ]
> }
> ```

This is not even valid JSON and therefore cannot be supported :)

> 
> or this case:
> 
> ```json
> {
>     "apiResources": [
>          null
>     ]
> }
> ```

Same situation as with first edge-case. We'll not support it. If you want to remove the array, you need to set `"apiResources": null`.

zongqichen:
> Nice summary of corner-cases, @zongqichen . I'll answer inline below:
> 
> > Hi @Fannon, I have some corner cases need to discuss:
> > 
> > 1. if set `null` in array, should I remove the whole array property? e.g:
> > 
> > ```json
> > {
> >     "apiResources": [
> >         {
> >             "ordId": "sap.sm:apiResource:SupplierService:v1",
> >             "partOfGroups": [
> >                 null,
> >                 "sap.cds:service:sap.test.cdsrc.sample:AdminService"
> >             ],
> >             "partOfPackage": "sap.sm:package:smDataProducts:v1"
> >         }
> >     ]
> > }
> > ```
> 
> To keep it simple (and stick with JSON Merge patch), we should always override arrays completely. Everything else will get complicated very quickly when ordering changes. Setting `null` inside an array is therefore invalid, as it would result in having a `null` entry in an array, which is not allowed in ORD.
> 
> > I should remove the ´partOfGroups´?
> > 
> > 2. if `null `set in default property array, should I leave it as empty or removed?
> > 
> > ```json
> > {
> >     "apiResources": [
> >         {null},
> >         {
> >             "ordId": "sap.sm:apiResource:SupplierService:v1",
> >             "partOfGroups": [
> >                 "sap.cds:service:sap.test.cdsrc.sample:AdminService"
> >             ],
> >             "partOfPackage": "sap.sm:package:smDataProducts:v1"
> >         }
> >     ]
> > }
> > ```
> 
> This is not even valid JSON and therefore cannot be supported :)
> 
> > or this case:
> > ```json
> > {
> >     "apiResources": [
> >          null
> >     ]
> > }
> > ```
> 
> Same situation as with first edge-case. We'll not support it. If you want to remove the array, you need to set `"apiResources": null`.

I read [JSON merge patch rfc7396](https://datatracker.ietf.org/doc/rfc7396/), I think it's slightly different with our use case. We should not always override arrays completely. If users want to update package localId for example,  they need to define following content, package ordId and localId they need to update:
```
{
    "packages": [
        {
            "ordId": "sap.sm:package:smDataProducts:v1",
            "localId": "smDataProducts"
        }
    ]
}
```
If we override completely, there are only two properties left. Our strategy is based on ordId, if no ordId mateched or provided, then extend it.

__tests__/__snapshots__/ordCdsrc.test.js.snap
```diff
@@ -5,11 +5,13 @@ exports[`Tests for default ORD document when .cdsrc.json is present Successfully
   "$schema": "https://sap.github.io/open-resource-discovery/spec-v1/interfaces/Document.schema.json",
   "apiResources": [
     {
-      "apiProtocol": "odata-v4",
-      "description": "AdminService",
+      "apiProtocol": "odata-v5",
```
Fannon, line 8: Why odata-v5? This would be an invalid choice / ORD document.

__tests__/__snapshots__/ordCdsrc.test.js.snap
```diff
@@ -5,11 +5,13 @@ exports[`Tests for default ORD document when .cdsrc.json is present Successfully
...
+      "description": "new AdminService",
       "entityTypeMappings": [
         {
-          "entityTypeTargets": [],
+          "entityTypeTargets": [
+            "new.entityType.target",
```
Fannon, line 13: This is also not compliant to ORD spec, it's actually an object with `ordId` of the target entity type


```suggestion
            {"ordId": ""sap.odm:entityType:BusinessPartner:v1"},
```

__tests__/__snapshots__/ordCdsrc.test.js.snap
zongqichen: I was thinking place a randomly value to distinguish test result, no special meaning. However, I have changed it in the coming commit.

__tests__/__snapshots__/ordCdsrc.test.js.snap
zongqichen: In the [document](https://pages.github.tools.sap/CentralEngineering/open-resource-discovery-specification/spec-v1/interfaces/document#entity-type-mapping_entitytypetargets), it shows the type is Array<[Entity Type Target (ORD ID)](https://pages.github.tools.sap/CentralEngineering/open-resource-discovery-specification/spec-v1/interfaces/document#entity-type-target-ord-id) | [Entity Type Target (Correlation ID)](https://pages.github.tools.sap/CentralEngineering/open-resource-discovery-specification/spec-v1/interfaces/document#entity-type-target-correlation-id)>. Is there update?

lib/extendOrdWithCustom.js
```diff
@@ -0,0 +1,86 @@
...
+const CONTENT_MERGE_KEY = 'ordId'
+
+function hasStringConflictsWithCdsrcJson(key) {
+    if (key in cds.env[ORD_ENV_KEY]) {
+        console.error(`Detect ord custom content has conflict with .cdsrc.json: ${key},
+            please resolve it, use .cdsrc.json value.`);
+        return true;
+    }
+    return false;
+}
```
Fannon, line 15: Not sure if we really need this. What you have in the `ord` object in .cdsrc file is not an ORD document and should not be merged anyway. We just merge the generated ORD content with the override content that is provided in the linked file.

lib/extendOrdWithCustom.js
```diff
@@ -0,0 +1,86 @@
...
+
+function handleStringConflicts(key) {
+    if (hasStringConflictsWithCdsrcJson(key)) {
+        return;
+    }
+    console.error(`Detect ord custom content has conflict with default value: ${key},
+        please define it in .cdsrc.json.`);
```
Fannon, line 22: Why would we need to run into this condition? Isn't the point of custom ORD content / overrides to be able to override default values?

swennemers: dead code?

__tests__/unittest/__snapshots__/extendOrdWithCustom.test.js.snap
```diff
@@ -0,0 +1,96 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`extendOrdWithCustom extendCustomORDContentIfExists should add new ord resources that are not supported by cap framework 1`] = `
```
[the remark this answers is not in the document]
zongqichen: It's generated by Jest, there is no document explained it. From the [code](https://github.com/jestjs/jest/blob/main/packages/jest-snapshot/src/State.ts#L206), I guess they want to avoid same test name override.

lib/extendOrdWithCustom.js
zongqichen: I want to warn users if they set top-level string property in customOrdFile, which won't be merged. Maybe give the function a better name? Or we don't need to handle this case and remove it?

Fannon: Yes, maybe the issue is then just the name of the function (which states cdsrc file as the source).

zongqichen: Good point, I also have an idea that we need to set proper logger. I'm not familiar with javascript logger, and I'm surprised that there is no buildin logger package inside nodejs. I found that cds providing a [logger ](https://cap.cloud.sap/docs/node.js/cds-log). What's your opinion?

swennemers (approved):

lib/extendOrdWithCustom.js
```diff
@@ -0,0 +1,68 @@
...
+        }
+    }
+    return cleanNullProperties(Object.values(destObj));
+}
+
+function compareAndHandleCustomORDContentWithExistingContent(ordContent, customORDContent, logger) {
+    const clonedOrdContent = structuredClone(ordContent);
+    for (const key in customORDContent) {
+        const propertyType = typeof customORDContent[key];
+        if (propertyType !== 'object' && propertyType !== 'array') {
```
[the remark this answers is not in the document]
swennemers: @aramovic79 , these are good comments. 
One of the use cases for this overwrite as well as also the @ORD.Extension is to allow a developer being ahead of the plugins Support for ORD. If we restrict the properties, we can't add things not yet available in the plug-in. But you are right, this comes with the risk of an broken ORD document. We can mitigate the risk with the validator and in case we want to experiment with new ORD features, we might even want the chance to do that.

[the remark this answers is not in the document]
zongqichen: Thanks, I link this comment to the https://github.com/cap-js/ord/issues/65, lets fix it in the following pr.

Issue #64, opened 2024-10-01
Feat: More flexible @ORD.Extension mechanism


Fannon:
The current implementation of `@ORD.Extension` overrides has a fixed number of ORD properties that it supports. We would need to constantly maintain this. It would therefore be better to implement the override / template logic completely generic.

The end result should be validated with the API Metadata Validator / or against the ORD JSON Schema (we could add the ORD Spec as an internal dependency if we want that) anyway.

Here is an example, from `template.js`:

https://github.com/cap-js/ord/blob/main/lib/templates.js#L223-L250

Instead we should do something like that (pseudo-code):

```js
let obj = {

        ordId: `${fReplaceSpecialCharacters(
          global.namespace
        )}:apiResource:${srvName}:v1`,
        title:
          srvDefinition["@title"] ??
          srvDefinition["@Common.Label"] ??
          srv,
        shortDescription: srv ?? 'No short description provided',
        description:
          srvDefinition["@Core.Description"] ?? 'No description provided',
       ...

       // now generically merge in rest and override the defaults above:
        ...ordExtensions,
      };
```

Fannon:
> let's take the validation out of this issue, even without ORD.Extension you can create strange ORD. Do you agree @Fannon ?

Agree. This would be a different issue if we want to add it in.

Issue #57, opened 2024-09-04
Variable name pattern


zongqichen:
I notice there are some variables named with special pattern, e.g. fReadORDExtensions, aServices, aODMEntity, having prefix like f and a. I want to discuss that, where does this pattern coming from? Do we have chance to get rid of it? Or we have to use it because it meets the requirements of CAP framework?

I briefly look other repos under cap-js, such as cds-typer and cds-dbs, there is no such naming pattern. So I want to listen others opinion.

Fannon:
I would be in favor of removing this "hungarian" notation. With an IDE and a compiler / language server, we already know what type a variable is. Adding types to variable names is therefore not needed and just adds noise, imho.
I would therefore just go with the typical JavaScript conventions as most open source projects, which is variable names in lowerCaseCamelCase that express the meaning (not the technical type).

ductaily:
Specifically `fReadORDExtensions` I would define functions using the `function` keyword. It is much more clear. If there is a reason why functions must be defined using const then it is a different story. But I am not aware of any. If anyone knows please correct me.

zongqichen:
> Specifically `fReadORDExtensions` I would define functions using the `function` keyword. It is much more clear. If there is a reason why functions must be defined using const then it is a different story. But I am not aware of any. If anyone knows please correct me.

I'm also confused about function creation in javascript, there are ways like statement, expression and arrow function. From my side, they are slightly different, but the purpose is same which is improving code readability and flexibility. Sometimes, expression and arrow functions are opposite, reduced the readability. I would propose that we use statement function as high priority but not strictly forbidden other ways to create function.

Fannon:
https://dev.to/skinnypetethegiraffe/function-or-const-which-do-you-prefer-typescriptjavascript-apa

States it well. Reasons for going with `function `:

- The intent is clear and concise, so that readers quickly differentiate between variables and functions. The faster you know what is going on, the better right?
- Function hoisting, meaning it lets you use a function before you declare it in your code, where with const it will be undefined until it has been reached in the execution.
- Also hoisting issues related to using function in bundling tools is no longer the issue it once was
- Named functions show up in stack trace messages, which is handy

PR #71, opened 2024-10-10
Fix: replace global variable and refactor code

zongqichen:


swennemers (approved):

lib/ord.js
```diff
@@ -176,68 +132,68 @@ const fGetEventResources = (csn, global, packageIds) => {
...
+        policyLevel: getPolicyLevel(appConfig),
+        description: getDescription(appConfig),
+        products: getProducts(appConfig),
+        groups: getGroups(linkedCsn, appConfig),
     };
     if (
-        fGetAPIResources(linkedCsn, global).length > 0 &&
-        fGetEventResources(linkedCsn, global).length > 0
+        getAPIResources(linkedCsn, appConfig).length > 0 &&
+        getEventResources(linkedCsn, appConfig).length > 0
```
aramovic79, line 172:
```suggestion
        getAPIResources(linkedCsn, appConfig).length &&
        getEventResources(linkedCsn, appConfig).length
```

lib/ord.js
```diff
@@ -176,68 +132,68 @@ const fGetEventResources = (csn, global, packageIds) => {
...
 
     // check if oReturn.packages is defined and extract the ordId from the packages and store in a set
     let packageIds = new Set();
-    if (oReturn.packages) {
-        oReturn.packages.forEach((pkg) => {
+    if (ordDocument.packages) {
+        ordDocument.packages.forEach((pkg) => {
             packageIds.add(pkg.ordId);
         });
     }
```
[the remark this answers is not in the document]
aramovic79: Another suggestion would be:
```
const packageIds = ordDocument.packages?.map((pkg) => pkg.ordId) || [];
```

aramovic79 (commented):
I've reported a few findings.
cc: @zongqichen @Fannon @swennemers

lib/ord.js
```diff
@@ -82,70 +77,37 @@ const fInitializeGlobal = (csn) => {
...
 
-/**
- * Retrieves the packages.
- * Hierarchy to check data: cdsrc.json > defaults.
- * @returns {Array<object>} The packages array.
- */
-const fGetPackages = (policyLevel, global) =>
-    global.env?.packages || global.aEvents.length
-        ? defaults.packages(global.appName, policyLevel, global.ordNamespace)
+const getPackages = (policyLevel, appConfig) =>
```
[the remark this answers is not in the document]
zongqichen: Agree

lib/ord.js
```diff
@@ -176,68 +132,68 @@ const fGetEventResources = (csn, global, packageIds) => {
...
+    let ordDocument = {
         $schema:
             "https://sap.github.io/open-resource-discovery/spec-v1/interfaces/Document.schema.json",
         openResourceDiscovery: "1.9",
-        policyLevel: fGetPolicyLevel(global),
-        description: fGetDescription(global),
-        products: fGetProducts(global),
-        groups: fGetGroups(linkedCsn, global),
+        policyLevel: getPolicyLevel(appConfig),
+        description: getDescription(appConfig),
```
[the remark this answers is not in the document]
zongqichen: I would like keep the private function. It makes the main function more readable. And we can set log or extend in the private function if needed

lib/ord.js
```diff
@@ -176,68 +132,68 @@ const fGetEventResources = (csn, global, packageIds) => {
...
-            global,
+            appConfig,
             packageIds
         )
     );
 };
 
 module.exports = (csn) => {
     const linkedCsn = cds.linked(csn);
-    Object.assign(global, fInitializeGlobal(linkedCsn));
```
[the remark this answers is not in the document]
aramovic79: The "problematic" flow would be:
1. We define appConfig and use appConfig in some function as an input parameter
2. We change property e.g. aServices in one of the functions called in the sequence
3. We use appConfig in some of the next functions in the sequence where we again check aServices for some criteria which is assuming that appConfig.aServices has the original value, BUT we filltered-out some values in the step 2
I'm saying it is the case right now, I'm stating it is a bad practice and makes code harder for maintenance because you don't know if in the future there will be some new functions added that are using appConfig and assuming it is unchanged...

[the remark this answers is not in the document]
zongqichen: Thanks, I get your point, but I'm not clear why do we need to assume or consider the appConfig has the original value? I think this question is key whether we should use cloned value. And I also want to know why it's bad practice? In some functions, I also use cloned [object](https://github.com/cap-js/ord/blob/main/lib/extendOrdWithCustom.js#L24). From my side, cloned object is expensive. We should use it when it really needed.

lib/ord.js
```diff
@@ -82,70 +77,37 @@ const fInitializeGlobal = (csn) => {
...
+const getProducts = (appConfig) =>
+    appConfig.env?.products || defaults.products(appConfig.packageName);
+
+const getGroups = (csn, appConfig) => {
+    return appConfig.aServices
         .map((srv) =>
-            fCreateGroupsTemplateForService(srv, csn.definitions[srv])
+            createGroupsTemplateForService(srv, csn.definitions[srv], appConfig)
         )
         .filter((resource) => resource !== null && resource !== undefined);
```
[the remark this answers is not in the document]
zongqichen: Thanks, I will use Boolean(x) instead

lib/templates.js
```diff
@@ -236,106 +181,63 @@ const fCreateAPIResourceTemplate = (serviceName, serviceDefinition, global, pack
...
-  if (packageIds instanceof Set) {
-    const packageArray = Array.from(packageIds);
-
-    if (apiOrEvent === "api") {
-      const apiPackage = packageArray.find((pkg) => pkg.includes("-api"));
-      if (apiPackage) return apiPackage;
-    } else if (apiOrEvent === "event") {
-      const eventPackage = packageArray.find((pkg) => pkg.includes("-event"));
-      if (eventPackage) return eventPackage;
+    if (packageIds instanceof Set) {
```
zongqichen, line 223: it should be removed

aramovic79:
Putting two cents:

enums vs objects in javascript

Enum can be defined as:
```
    const RESOURCE_TYPE = Object.freeze(
    {
        "api": "api",
        "event": "event"
    });
```
Please pay attention on **Object.freeze**. It serves to disallow changing of the object which at the end causes that RESOURCE_TYPE acts as an enum.

aramovic79:
I think that the code: 
```
function _getPackageID(namespace, packageIds, apiOrEvent) {
    if (packageIds instanceof Set) {
        const packageArray = Array.from(packageIds);

        if (apiOrEvent === "api") {
            const apiPackage = packageArray.find((pkg) => pkg.includes("-api"));
            if (apiPackage) return apiPackage;
        } else if (apiOrEvent === "event") {
            const eventPackage = packageArray.find((pkg) => pkg.includes("-event"));
            if (eventPackage) return eventPackage;
        }

        return packageArray.find(pkg => pkg.includes(namespace));
    }
}
```

can be written this way:

```
function _getPackageID(namespace, packageIds, apiOrEvent) {
    return packageIds.find((pkg) => pkg.includes("-" + apiOrEvent)) || packageIds.find((pkg) => pkg.includes(namespace));
}

```
where I think that instead of ```apiOrEvent``` we should name this input parameter like ```resourceType```

lib/templates.js
zongqichen: https://github.com/cap-js/ord/issues/73

lib/templates.js
```diff
@@ -26,28 +26,28 @@ const fReadORDExtensions = (srv) =>
  * @returns {Array} An array containing paths and it's kind.
  */
 
-const fGeneratePaths = (srv, srvDefinition) => {
-  const srvObj = { name: srv, definition: srvDefinition };
-  const protocols = cds.service.protocols;
+const generatePaths = (srv, srvDefinition) => {
```
aramovic79, line 29: This is a private function, right? Maybe to name it `_generatePath`?
Also, when defining functions, on some places we use `function` and on the others `const`. Do we follow some rule when to use what? I couldn't find any regularity, it seems it was randomly/arbitrarily decided.

lib/templates.js
```diff
@@ -7,16 +7,16 @@ const cds = require("@sap/cds");
...
-  Object.entries(srv)
-    .filter(([key]) => key.startsWith("@ORD.Extensions."))
-    .reduce(
-      (ordExtensions, [key, value]) => ({
-        ...ordExtensions,
-        [key.slice("@ORD.Extensions.".length)]: value,
-      }),
-      {}
-    );
+const readORDExtensions = (srv) =>
```
aramovic79, line 10: We are calling this function in multiple places in a sequence when generating ord document and every time we have the same input parameter - serviceDefinition. Why not to call this function once when starting generating ORD document and then pass the result as an input parameter of all subsequent functions: _getGroups, _getApiResources, _getEventResources... ?

aramovic79 (approved):

lib/templates.js
aramovic79: image

swennemers: let's close this PR and fix this as part of the next PR #64

lib/templates.js
zongqichen: As https://github.com/cap-js/ord/issues/57 discussed, we should use function

PR #72, opened 2024-10-14
Update extendOrdWithCustom.js

RamIndia:
Minor condition check

### DIFF
--- a/lib/extendOrdWithCustom.js
+++ b/lib/extendOrdWithCustom.js
@@ -47,7 +47,7 @@ function compareAndHandleCustomORDContentWithExistingContent(ordContent, customO
 }
 
 function getCustomORDContent(appConfig) {
-    if (appConfig.env.customOrdContentFile) {
+    if (appConfig.env?.customOrdContentFile) {
         const customORDContent = require(path.join(cds.root, appConfig.env.customOrdContentFile));
         return customORDContent;
     }


lib/extendOrdWithCustom.js
```diff
@@ -47,7 +47,7 @@ function compareAndHandleCustomORDContentWithExistingContent(ordContent, customO
 }
 
 function getCustomORDContent(appConfig) {
-    if (appConfig.env.customOrdContentFile) {
+    if (appConfig.env?.customOrdContentFile) {
```
zongqichen, line 50: Hi @RamIndia, why do you need this question mark? You can see the [code](https://github.com/cap-js/ord/blob/main/lib/ord.js#L74), there is always env in appConfig.

aramovic79 (commented):
I would suggest first to review/approve the PR 71 and then to include this change.

lib/extendOrdWithCustom.js
Fannon: Is this really given? What happens if there's no `ord` dictionary in the `.cdsrc` file? But if you want to fall back to a non-null value, you could do the "default" fallback in the linked file like:

```js
env: cds.env["ord"] || {},
```

zongqichen: Good point, I create an issue: https://github.com/cap-js/ord/issues/74

zongqichen (approved):
I test the case without ord in .cdsrc.json, this fix is necessary to prevent error raised.

Issue #77, opened 2024-10-23
Filter out metatata where visibility=internal or =private


Fannon:
So far this plugin does not support protection of the metadata. To prevent accidental leaking of internal information, we should filter out metadata that has `visibility: internal` or `visibility: private`. 

We should also add a disclaimer to the plugin that by installing, the metadata will be openly accessible and that protection of the metadata is not yet supported, but planned.

zongqichen:
Hi @Fannon, should we consider which environment we can stop filtering? For example, when `env=production`, we filter out, but in dev or local environment, we skip filter? Otherwise, is same as we remove it

Fannon:
> Hi @Fannon, should we consider which environment we can stop filtering? For example, when `env=production`, we filter out, but in dev or local environment, we skip filter? Otherwise, is same as we remove it

No, I think this should depend on whether the security / auth for ORD metadata is given. If there is just "open" access, we can only expose public metadata. 

Only once we support protected metadata access, we can really expose internal / private - otherwise we would have a security issue.

zongqichen:
> > Hi @Fannon, should we consider which environment we can stop filtering? For example, when `env=production`, we filter out, but in dev or local environment, we skip filter? Otherwise, is same as we remove it
> 
> No, I think this should depend on whether the security / auth for ORD metadata is given. If there is just "open" access, we can only expose public metadata.
> 
> Only once we support protected metadata access, we can really expose internal / private - otherwise we would have a security issue.

Based on the [docs](https://pages.github.tools.sap/CentralEngineering/open-resource-discovery-specification/spec-v1/interfaces/document#api-resource_visibility):
```
"internal": Resources are accessible to other applications within the same organization.
```
The current implementation will not distinguish `internal` and `private`, they will be filtered out at all. Is that correct?

Fannon:
@zongqichen : Yes, `internal` and `private` both needs to be protected, so you filter both of them out.

PR #79, opened 2024-10-23
Add disclaimer that metadata access is currently open access

Fannon:
Relates #77 

I think we really should point this out explicitly and mention that protected metadata access is not (yet) available.


zongqichen (approved):
LGTM

swennemers (approved):

PR #80, opened 2024-10-23
Update changelog and version

zongqichen:


aramovic79 (commented):
I wrote a proposal how the CHANGELOG.md should/can look like. It is generated by GitHub.
I also don't mind to keep Zongqi's proposal.
@RamIndia : What is the practice you are following?

aramovic79 (approved):

PR #82, opened 2024-10-24
feat: Filter out internal and private metadata

aramovic79:


lib/templates.js
```diff
@@ -193,7 +193,7 @@ const createEventResourceTemplate = (serviceName, eventDefinition, appConfig, pa
         releaseStatus: "active",
         partOfPackage: _getPackageID(appConfig.ordNamespace, packageIds, 'event'),
         partOfGroups: [_getGroupID(serviceName, defaults.groupTypeId, appConfig)],
-        visibility: "public",
+        visibility: "public", // TODO: visibility is always set to public
```
aramovic79, line 196: @Fannon @swennemers @zongqichen : As of now, the visibility of a resource is always set to `public`. 
I guess this should be changed to consider the (if existing) value of the `visibility` property in the `@ORD.Extensions`, like in the following example: 
`annotate ProcessorService with @ORD.Extensions: {
    title           : 'This is Processor Service title',
    shortDescription: 'short description for Processor Service',
    **visibility      : 'public',**
    extensible      : {supported: 'no'}
};`
Only then it would make sense to apply the filtering.
Am I correct?

Fannon: @aramovic79 : yes, via `ORD.Extensions: { visibility: 'internal' }` you can already set it to internal. Once you do that, filtering would apply.

xmpl/srv/services.cds
```diff
@@ -9,7 +9,7 @@ namespace sap.capire.incidents;
 annotate ProcessorService with @ORD.Extensions: {
     title           : 'This is Processor Service title',
     shortDescription: 'short description for Processor Service',
-    visibility      : 'public',
+    visibility      : 'private',
```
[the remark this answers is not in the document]
aramovic79: Sure, the pull request is just a draft, I've pushed initial changes. This change was added for internal testing, will be reverted.

lib/templates.js
```diff
@@ -173,11 +176,11 @@ const createAPIResourceTemplate = (serviceName, serviceDefinition, appConfig, pa
 
             ...ordExtensions,
         };
-
-        apiResources.push(obj);
+        // visibility can also be set as ORD extension, in which case it overwrites the default visibility
+        if (obj.visibility === RESOURCE_VISIBILITY.public) apiResources.push(obj);
     });
 
-    if (apiResources.length > 0) return apiResources;
```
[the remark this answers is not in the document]
aramovic79: This is part of the changes done in ord.js as well; and this is just an initial commit, there are much more to come, including test-changes and new tests as well.

__tests__/__snapshots__/ordCdsrc.test.js.snap
```diff
@@ -4,55 +4,6 @@ exports[`Tests for default ORD document when .cdsrc.json is present Successfully
 {
   "$schema": "https://sap.github.io/open-resource-discovery/spec-v1/interfaces/Document.schema.json",
   "apiResources": [
-    {
```
aramovic79, line 7: @zongqichen : Since the `visibility` of AdminService in the __tests__\bookshop\srv\admin-service.cds is set now to `private`, none of it's resources(apiResources, evenResources) will be present in generated ORD Document.
The rule is: `private` and `internal` resources are hidden, only the `public` ones will be shown in ORD Doc.

__tests__/__snapshots__/ordCdsrc.test.js.snap
```diff
@@ -105,6 +56,21 @@ exports[`Tests for default ORD document when .cdsrc.json is present Successfully
       "version": "1.0.0",
       "visibility": "public",
     },
+    {
```
aramovic79, line 59: @zongqichen : This is "coming" from __tests__\bookshop\ord\custom.ord.json. Not sure if we want to inject this into the resulting ORD Document. What is your opinion?

__tests__/__snapshots__/ordCdsrc.test.js.snap
```diff
@@ -144,42 +110,6 @@ exports[`Tests for default ORD document when .cdsrc.json is present Successfully
   ],
   "description": "this is my custom description",
   "eventResources": [
-    {
-      "description": "CAP Event resource describing events / messages.",
```
aramovic79, line 148: @zongqichen : Similar as for apiResources. Pls check the comment from above.

__tests__/__snapshots__/ordPackageJson.test.js.snap
```diff
@@ -4,56 +4,6 @@ exports[`Tests for default ORD document when .cdsrc.json is not present Successf
...
-          ],
-          "mediaType": "application/xml",
-          "type": "edmx",
-          "url": "/.well-known/open-resource-discovery/v1/api-metadata/AdminService.edmx",
-        },
-      ],
-      "shortDescription": "short description for test AdminService",
-      "title": "This is test AdminService title",
-      "version": "2.0.0",
-      "visibility": "private",
```
aramovic79, line 55: @zongqichen : Similar as above, the `private` resource should not be shown in the resulting ORD Document.

__tests__/ordCdsrc.test.js
```diff
@@ -39,12 +39,16 @@ describe("Tests for default ORD document when .cdsrc.json is present", () => {
 
         test("PartOfPackage values are valid ORD IDs ", () => {
             for (const apiResource of document.apiResources) {
+                if (!apiResource.partOfPackage) continue;
```
aramovic79, line 42: @zongqichen : In case of the `injected` resource(the one given in `__tests__\bookshop\ord\custom.ord.json`), where no `partOfPackage` is stated, this check should be skipped.

__tests__/ordCdsrc.test.js
```diff
@@ -39,12 +39,16 @@ describe("Tests for default ORD document when .cdsrc.json is present", () => {
...
+
                 expect(apiResource.partOfPackage).toMatch(PACKAGE_ID_REGEX);
             }
         });
 
         test("The partOfPackage references an existing package", () => {
             for (const apiResource of document.apiResources) {
+                if (!apiResource.partOfPackage) continue;
```
aramovic79, line 50: @zongqichen : The same as above.

__tests__/unittest/__snapshots__/templates.test.js.snap
```diff
@@ -54,91 +54,18 @@ exports[`templates createAPIResourceTemplate should create API resource template
 `;
 
 exports[`templates createEventResourceTemplate should create event resource template correctly 1`] = `
-{
```
aramovic79, line 57: @zongqichen : This snapshot is re-generated(by using `npx jest -u`. Does it seem correct to you? Not sure if this is the command that should be used or I've missed something?

__tests__/unittest/extendOrdWithCustom.test.js
```diff
@@ -45,9 +44,13 @@ describe('extendOrdWithCustom', () => {
...
             const ordContent = {};
+            const warningSpy = jest.spyOn(console, 'warn');
             prepareTestEnvironment({ namespace: "sap.sample" }, appConfig, 'testCustomORDContentFileThrowErrors.json');
-            const result = extendCustomORDContentIfExists(appConfig, ordContent, logger);
-            expect(logger.warn).toHaveBeenCalledTimes(3);
+            const result = extendCustomORDContentIfExists(appConfig, ordContent);
+
+            expect(warningSpy).toHaveBeenCalledTimes(3);
+            expect(warningSpy).toHaveBeenCalledWith('Mocked warning');
+
```
aramovic79, line 53: @zongqichen : This way we can also test if the `cds.log.warn` is called in function `compareAndHandleCustomORDContentWithExistingContent`. 
Benefit of this way of testing is that we do not need to propagate the `logger` through 3 functions. So, this way we don't need parameter `logger` is the following functions:
```
extendCustomORDContentIfExists
compareAndHandleCustomORDContentWithExistingContent
```

__tests__/unittest/templates.test.js
```diff
@@ -43,9 +43,7 @@ describe('templates', () => {
         it('should create API resource template correctly', () => {
             const serviceName = 'MyService';
             const srvDefinition = linkedModel
-            const packageIds = new Set()
-            packageIds.add('sap.test.cdsrc.sample:package:test-event:v1');
-            packageIds.add('sap.test.cdsrc.sample:package:test-api:v1');
+            const packageIds = ['sap.test.cdsrc.sample:package:test-event:v1', 'sap.test.cdsrc.sample:package:test-api:v1'];
```
aramovic79, line 46: @zongqichen : Using `Array` instead of `Set` so we don't need to convert the `Set` into `Array` in different places, e.g., in the `_getPackageID` function.

lib/constants.js
```diff
@@ -1,33 +1,39 @@
...
+const ORD_EXTENSIONS_PREFIX = "@ORD.Extensions.";
+
+const ORD_RESOURCE_TYPE = Object.freeze({
+    service: "service",
+    entity: "entity",
+    event: "event",
+    api: "api",
+});
+
+const RESOURCE_VISIBILITY = Object.freeze({
```
aramovic79, line 22: @zongqichen : Minor change - just sorted the constants alphabetically. This is recommended by the linter.

lib/extendOrdWithCustom.js
```diff
@@ -28,12 +28,12 @@ function patchGeneratedOrdResources(destinationObj, sourceObj) {
...
 }
 
-function compareAndHandleCustomORDContentWithExistingContent(ordContent, customORDContent, logger) {
+function compareAndHandleCustomORDContentWithExistingContent(ordContent, customORDContent) {
     const clonedOrdContent = structuredClone(ordContent);
     for (const key in customORDContent) {
         const propertyType = typeof customORDContent[key];
         if (propertyType !== 'object' && propertyType !== 'array') {
-            logger.warn('Found ord top level primitive ord property in customOrdFile:', key, ', please define it in .cdsrc.json.');
+            log('ord-plugin').warn('Found ord top level primitive ord property in customOrdFile:', key, ', please define it in .cdsrc.json.');
```
aramovic79, line 36: @zongqichen : Explanation given above(no need to propagate input parameter `logger`).

lib/ord.js
```diff
@@ -1,25 +1,22 @@
...
+const cds = require("@sap/cds");
+const defaults = require("./defaults");
 const { extendCustomORDContentIfExists } = require('./extendOrdWithCustom');
-const { ORD_RESOURCE_TYPE } = require('./constants');
-const { getRFC3339Date } = require('./date')
-
-const logger = cds.log('ord-plugin');
+const { getRFC3339Date } = require('./date');
+const _ = require("lodash");
+const { join } = require("path");
```
aramovic79, line 13: @zongqichen : Sorted alphabetically.

lib/ord.js
```diff
@@ -100,15 +97,14 @@ const _getPackages = (policyLevel, appConfig) =>
 
 const _getAPIResources = (csn, appConfig, packageIds) => {
     return appConfig.serviceNames
-        .flatMap((serviceName) => createAPIResourceTemplate(serviceName, csn.definitions[serviceName], appConfig, packageIds) || [])
-        .filter((resource) => !!resource);
+        .flatMap((serviceName) => createAPIResourceTemplate(serviceName, csn.definitions[serviceName], appConfig, packageIds))
```
aramovic79, line 100: @zongqichen : Function `createAPIResourceTemplate` anyways returns an empty array if there are no(public) service-resources.

lib/ord.js
```diff
@@ -100,15 +97,14 @@ const _getPackages = (policyLevel, appConfig) =>
...
 };
 
 const _getEventResources = (csn, appConfig, packageIds) => {
     if (appConfig.events.length === 0) return [];
     return appConfig.serviceNames
         .filter((serviceName) => appConfig.events.some((eventName) => eventName.startsWith(serviceName)))
-        .flatMap((serviceName) => createEventResourceTemplate(serviceName, csn.definitions[serviceName], appConfig, packageIds) || []);
+        .flatMap((serviceName) => createEventResourceTemplate(serviceName, csn.definitions[serviceName], appConfig, packageIds));
```
aramovic79, line 107: @zongqichen : The same applies here as for apiResources.

lib/ord.js
```diff
@@ -151,9 +147,9 @@ function createDefaultORDDocument(linkedCsn, appConfig) {
 }
 
 function extractPackageIds(ordDocument) {
-    const packageIds = new Set();
+    const packageIds = [];
     if (ordDocument.packages) {
-        ordDocument.packages.map((pkg) => packageIds.add(pkg.ordId));
+        ordDocument.packages.map((pkg) => packageIds.push(pkg.ordId));
```
aramovic79, line 152: @zongqichen : I though there is no need for using `Set`; instead, we should use `Array` data-structure.

lib/templates.js
```diff
@@ -173,31 +177,40 @@ const createAPIResourceTemplate = (serviceName, serviceDefinition, appConfig, pa
...
+    const ordExtensions = readORDExtensions(serviceDefinition);
+    if (!!ordExtensions.visibility && ordExtensions.visibility !== RESOURCE_VISIBILITY.public) return [];
+    return [{
         ordId: `${appConfig.ordNamespace}:eventResource:${serviceName}:v1`,
         title:
-            eventDefinition["@title"] ??
-            eventDefinition["@Common.Label"] ??
+            serviceDefinition["@title"] ??
+            serviceDefinition["@Common.Label"] ??
             `ODM ${appConfig.appName.replace(/[^a-zA-Z0-9]/g, "")} Events`,
```
aramovic79, line 203: @zongqichen : This is definitely a whole service and not only an event definition. That's why I've renamed it into `serviceDefinition`.

lib/templates.js
```diff
@@ -213,14 +226,13 @@ const createEventResourceTemplate = (serviceName, eventDefinition, appConfig, pa
...
 };
 
 function _getPackageID(namespace, packageIds, resourceType) {
-    //TODO: without this check, it will be failed in ordPackageJson.test.js, need to check why it is failing
-    if (!(packageIds instanceof Set)) return;
-    const packageIdsArray = Array.from(packageIds);
-    return packageIdsArray.find((pkg) => pkg.includes("-" + resourceType)) || packageIdsArray.find((pkg) => pkg.includes(namespace));
+    if (!packageIds) return;
+
+    return packageIds.find((id) => id.includes("-" + resourceType)) || packageIds.find((id) => id.includes(namespace));
```
aramovic79, line 235: @zongqichen : As explained above, no need to check if the `packageIds` is of type `Set` because it can be an `Array` or `undefined`.

lib/constants.js
zongqichen: How do you do it? I like it, in goland, there is a linter can achieve. Is it an extension in vscode which we could auto-save? We should put it into vscode settings, which any developer can follow same way

lib/extendOrdWithCustom.js
```diff
@@ -1,5 +1,5 @@
-const path = require("path");
-const cds = require("@sap/cds");
+const { root, log } = require("@sap/cds");
+const { join } = require("path");
```
zongqichen, line 2: From my side, I would prefer `path.join`. When I read `join` independently, I need more context what is about, could be string.join, list.join and so on. `path.join` can provide the context, what do you think?

lib/extendOrdWithCustom.js
zongqichen: From my opinion, the central logger object is needed. The reasons are:
- if we want to change logger name: 'ord-plugin' -> 'ord', we have to maintain it in all different places
- if we have special configuration to logger, e.g set level, env, or save debug msg into file, central logger can achieve it
- I guess put logger as argument won't spend extra costs? 

And I have an issue wants to put central logger to all necessary places. I learnt it from previous golang project, which we define the logger object at the begging and spread it to all places. I'm not sure how should javascript best practice look like, what's your idea?

lib/ord.js
```diff
@@ -1,25 +1,22 @@
...
+const _ = require("lodash");
+const { join } = require("path");
 
 const initializeAppConfig = (csn) => {
-    let packagejsonPath = path.join(cds.root, 'package.json')
+    let packageJsonPath = join(cds.root, 'package.json')
     let packageJson;
-    if (exists(packagejsonPath)) {
-        packageJson = require(packagejsonPath);
+    if (cds.utils.exists(packageJsonPath)) {
```
zongqichen, line 18: This is good example! `cds.utils.exists` gives me the clear context. I wish `path.join` is as well.

lib/ord.js
```diff
@@ -31,8 +28,8 @@ const initializeAppConfig = (csn) => {
     const lastUpdate = getRFC3339Date();
     let odmEntity = [];
 
-    // namespace variable value if present in cdsrc.json take it there or else from package.json
-    //if cdsrc.json does not have applicationNamespace, then use just the namespace
+    // use "namespace" from package.json as fallback in case it's not given in cdsrc.json
+    // if cdsrc.json does not have "applicationNamespace", use the "namespace"
```
zongqichen, line 32: We can remove it. When I read it, it can't help me have deeper understanding. And we have to maintain it when something changed

lib/ord.js
zongqichen: I'm thinking we should move all private functions like `_getApiResources` out of ord.js. Now we don't have unittest to protect them.

lib/templates.js
```diff
@@ -158,8 +162,8 @@ const createAPIResourceTemplate = (serviceName, serviceDefinition, appConfig, pa
                 DESCRIPTION_PREFIX + serviceName,
             version: "1.0.0",
             lastUpdate: appConfig.lastUpdate,
-            visibility: "public",
-            partOfPackage: _getPackageID(appConfig.ordNamespace, packageIds, "api"),
+            visibility: RESOURCE_VISIBILITY.public, // default visibility
```
zongqichen, line 165: The constant is great! However, the comment behind should be removed, no needed I guess?

lib/templates.js
```diff
@@ -173,31 +177,40 @@ const createAPIResourceTemplate = (serviceName, serviceDefinition, appConfig, pa
 
             ...ordExtensions,
         };
-
-        apiResources.push(obj);
+        // visibility can also be set as an ORD extension; in this case, it overwrites the default visibility
+        if (obj.visibility === RESOURCE_VISIBILITY.public) apiResources.push(obj);
```
zongqichen, line 181: I would suggest we move this logic to other place. e.g. ord.js. And create a function might be so called: `filterNotPublicObj()`. The reasons:
- apiResources and eventResources can use the same function
- I'm not sure in the future, should we consider we hide the information in production env, but output it in dev or test env?

lib/templates.js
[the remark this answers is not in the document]
zongqichen: We need unittest case to against it in templates.test.js. Make this change visible by test.

lib/templates.js
```diff
@@ -173,31 +177,40 @@ const createAPIResourceTemplate = (serviceName, serviceDefinition, appConfig, pa
...
+ * This is a template function to create Event Resource object for Event Resource Array.
+ * There can be only one event resource per service because all events are using the same protocol, they are always Cloud Events.
+ *
+ * @param {string} serviceName The name of the service.
+ * @param {object} serviceDefinition The definition of the service
+ * @returns {Array} An single-item array of objects for the Event Resources.
+ */
+const createEventResourceTemplate = (serviceName, serviceDefinition, appConfig, packageIds) => {
+    const ordExtensions = readORDExtensions(serviceDefinition);
+    if (!!ordExtensions.visibility && ordExtensions.visibility !== RESOURCE_VISIBILITY.public) return [];
```
zongqichen, line 197: Need unittest case against it and the logic is slightly different with apiResources?

lib/templates.js
```diff
@@ -173,31 +177,40 @@ const createAPIResourceTemplate = (serviceName, serviceDefinition, appConfig, pa
...
+            serviceDefinition['@description'] ?? serviceDefinition['@Core.Description'] ??
             "CAP Event resource describing events / messages.",
         version: "1.0.0",
         lastUpdate: appConfig.lastUpdate,
         releaseStatus: "active",
-        partOfPackage: _getPackageID(appConfig.ordNamespace, packageIds, 'event'),
+        partOfPackage: _getPackageID(appConfig.ordNamespace, packageIds, ORD_RESOURCE_TYPE.event),
         partOfGroups: [_getGroupID(serviceName, defaults.groupTypeId, appConfig)],
-        visibility: "public",
+        visibility: RESOURCE_VISIBILITY.public,//default value that can be overwritten by ORD extension
```
zongqichen, line 213: The comment is not needed

lib/templates.js
```diff
@@ -213,14 +226,13 @@ const createEventResourceTemplate = (serviceName, eventDefinition, appConfig, pa
         extensible: { supported: "no" },
 
         ...ordExtensions
-    };
+    }]
```
zongqichen, line 229: The list is needed? If it is, we also need unittest to protect it

__tests__/unittest/__snapshots__/templates.test.js.snap
zongqichen: Re-generated is fine, but we need to remove the visibility from extensions. Otherwise, we will lost the test. I would suggest add new testcase, e.g. `it(should filter out not public object)`

aramovic79: @zongqichen :
4 new assertions are added into the templates.test.js: for private and internal visibility of apiResources and apiEvents.
The check if both functions(getApiResources and getEventResources) are returning arrays are added as well.
Reminder: before, the function getEventResources was returning an object; this is wrong because it should return **single-item array**.
cc: @Fannon, @swennemers

lib/templates.js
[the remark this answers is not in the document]
aramovic79: Unit test cases added.

__tests__/__snapshots__/ordCdsrc.test.js.snap
aramovic79: As suggested, AdminService is now **public**, CatalogService is **private** and CinemaService is **internal**.

lib/constants.js
aramovic79: It is configuration in linter which is not part of the plugin but set in calesi-root.
This time I did it by using GH-Copilot. I find alphabetical sorting `require` and `constants` standing in the header pretty convenient. We can do that in a separate PR.

lib/ord.js
aramovic79: I agree. We should do that in a separate PR.

lib/ord.js
aramovic79: Good point! I will check now.

lib/templates.js
```diff
@@ -173,31 +177,40 @@ const createAPIResourceTemplate = (serviceName, serviceDefinition, appConfig, pa
...
-    const ordExtensions = readORDExtensions(eventDefinition);
-    return {
+/**
+ * This is a template function to create Event Resource object for Event Resource Array.
+ * There can be only one event resource per service because all events are using the same protocol, they are always Cloud Events.
+ *
+ * @param {string} serviceName The name of the service.
+ * @param {object} serviceDefinition The definition of the service
+ * @returns {Array} An single-item array of objects for the Event Resources.
+ */
```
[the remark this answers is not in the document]
aramovic79: @zongqichen : I think that the line 189 is needed because it's not obvious from the code that there can be only one event resource per service(because all the events are using the same protocol). This information seems useful for me.
@Fannon @swennemers : Do you agree to keep this doc comment?

lib/extendOrdWithCustom.js
Fannon: Yes, refactoring logging should be separate PR. 

If it's a factory we get, I assume we should have a singleton instance created for the entire plugin that we then re-use in different files?

lib/templates.js
Fannon: Everything here can be overwritten by ORDExtensions I assume? Maybe that comment should be moved to the "higher level"

aramovic79: Moved to the function doc comment.

lib/extendOrdWithCustom.js
aramovic79: Let's do that in a separate PR.

lib/templates.js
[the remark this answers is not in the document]
swennemers: @aramovic79 , @zongqichen - please align on the code style around comments.

lib/templates.js
swennemers: @aramovic79 , I assume you mean something different than @zongqichen . Having a consistent filter function like "isVisible(ordResource)" might be useful, esp. if we introduce additional logic around the filtering, e.g. if correctly authenticated, then also show internal and private resources... such additions can than be implemented in such a central function. 
Anyhow I wouldn't overthink it now, you can do such refactoring also later.

lib/ord.js
aramovic79: I will revert the change then. 👍

zongqichen (approved):
LGTM

Issue #58, opened 2024-09-09
multi-tenancy & extensibility aware ORD plugin


swennemers:
So far it is unclear if the ORD plugin works also in a multi-tenancy aware CAP app using the CAP-MTX features. See also https://github.com/SAP-samples/btp-cap-multitenant-saas.

@RoshniNaveenaS and @RamIndia, what are your suggestions to have a simple test case ensuring the plugin also works in a multi-tenant environment and will return tenant specific extensions?

RamIndia:
@swennemers : currently ORD plugin works with single tenant apps, multi-tenancy is not supported yet but it will be added later.

swennemers:
@RamIndia : what are the essentials missing to implement multi-tenancy support, that is essential for SAP SaaSs apps using the plugin?

RamIndia:
@swennemers we are still discussing about it in sync up's how do we proceed with it, we will update once we have a clarity.

zongqichen:
As confirmed from CAP team, cds.model can be supported to multitenancy and extensibility.

PR #83, opened 2024-10-25
feat: add custom index page to the ord-sample

aramovic79:


aramovic79:
Bellow is shown how the index page looks like now. 
As you can see in the code, the implemented changes have nothing to do with the ORD Plugin, but only with the example application(/xmpl).

image

swennemers:
Thx @aramovic79 to immediately jump on the issue. I would prefer a solution where the plugin can provide this snippet, I think, I saw some other plugins to contribute to the page once. @RamIndia , can you help with the extension points to hook in for that?

swennemers:
@aramovic79 & @zongqichen , have a look at https://www.npmjs.com/package/@sap/cds?activeTab=code /@sap/cds/app/index.js, the plugin should return its endpoint as a CAP service (although manually implemented), then it would appear here automatically. @RamIndia suggested to do the same to support the multi tenancy issue #58.

aramovic79:
> @aramovic79 & @zongqichen , have a look at https://www.npmjs.com/package/@sap/cds?activeTab=code /@sap/cds/app/index.js, the plugin should return its endpoint as a CAP service (although manually implemented), then it would appear here automatically. @RamIndia suggested to do the same to support the multi tenancy issue #58.

By looking at the code, it seems to me that the change should be implemented not even in @sap/cds but in @sap/cds-dk. @RamIndia : Am I right?
If so, not sure which change shall be done in the ORD Plugin so that @sap/cds(index.js) "sees" it's endpoint as a CAP service. I think that diving into the code with a debugger is not the right approach, we need someone from the CAP team to point us in the right direction.

xmpl/package.json
```diff
@@ -1,14 +1,22 @@
...
+    "devDependencies": {
+        "@cap-js/sqlite": "*"
+    },
+    "private": true,
+    "cds": {
+        "requires": {
+            "SAP ORD Service": {
+                "model": "@cap-js/ord/srv/ord-service",
+                "service": "OrdService"
+            }
```
swennemers, line 19: this should be part of the plugins package.json, shouldn't it?

xmpl/srv/ord-service.cds
swennemers, line 1: these files should per part of the plugin, not the xmpl

aramovic79:
Obsolete.

PR #89, opened 2024-11-04
Feat: add default lastUpdate and consumptionBundles

zongqichen:


lib/date.js
```diff
@@ -0,0 +1,18 @@
...
+    const day = String(now.getUTCDate()).padStart(2, '0');
+    const hours = String(now.getUTCHours()).padStart(2, '0');
+    const minutes = String(now.getUTCMinutes()).padStart(2, '0');
+    const seconds = String(now.getUTCSeconds()).padStart(2, '0');
+    const offsetHours = '01';
+    const offsetMinutes = '00';
+    const offsetSign = '+';
+
+    return `${year}-${month}-${day}T${hours}:${minutes}:${seconds}${offsetSign}${offsetHours}:${offsetMinutes}`;
+}
```
aramovic79, line 14:
```suggestion
function getRFC3339Date(includeOffset = true) {
    const now = new Date();
    const year = now.getUTCFullYear();
    const month = String(now.getUTCMonth() + 1).padStart(2, '0');
    const day = String(now.getUTCDate()).padStart(2, '0');
    const hours = String(now.getUTCHours()).padStart(2, '0');
    const minutes = String(now.getUTCMinutes()).padStart(2, '0');
    const seconds = String(now.getUTCSeconds()).padStart(2, '0');
    
    if (includeOffset) {
        const offsetHours = '01';
        const offsetMinutes = '00';
        const offsetSign = '+';
        return `${year}-${month}-${day}T${hours}:${minutes}:${seconds}${offsetSign}${offsetHours}:${offsetMinutes}`;
    } else {
        return `${year}-${month}-${day}T${hours}:${minutes}:${seconds}Z`;
    }
}
```

__tests__/unittest/date.test.js
```diff
@@ -0,0 +1,24 @@
+const { getRFC3339Date } = require('../../lib/date');
+
+const RFC3339_REGEX = /^((?:(\d{4}-\d{2}-\d{2})T(\d{2}:\d{2}:\d{2}(?:\.\d+)?))(Z|[+-]\d{2}:\d{2})?)$/;
```
aramovic79, line 3:
```suggestion
const RFC3339_REGEX = /^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}(?:\.\d+)?(?:Z|[+-]\d{2}:\d{2})$/;
```

aramovic79 (approved):
Two small suggestions added. Otherwise, looks good to me.

__tests__/unittest/date.test.js
zongqichen: Hi @aramovic79 , good catch. I try to use validator to check the lastUpdate without offset `2024-06-20T14:04:01` and it passed. So I'm curious, how do you know lastUpdate can be without offset or do you know somewhere use it as well? Thx

aramovic79: According to the RFC 3339 datetime-format definition, if there is no offset(so, the time-offset is 0), the letter **Z** should stand at the end, to show that the datetime is UTC. 
Regex should not "allow" a datetime without time-offset or letter Z at the end. 
If my proposed regex allows it, forget it. 😊

Fannon: This should match the date format in the ORD spec, but there we just refer to https://datatracker.ietf.org/doc/html/rfc3339#section-5.6 and validate it as a JSON Schema "format": "date". I'm not entirely sure what the regexp of JSON Schema dates looks like, but we should compare this to make sure we got the same validation and understanding.

zongqichen: Hi @Fannon , the unittest cases are taken from [rfc3339#section-5.8 examples](https://datatracker.ietf.org/doc/html/rfc3339#section-5.8). I think the current implementation should be fine, later we can use validator to check the correctness of ord json format.

Issue #86, opened 2024-10-28
Fix: add error handling when custom.ord.json doesn't exist


zongqichen:
If there is no custom.ord.json filed existed, there will be throw an error: `Cannot find module....`, we need to reflect more clear error message.

Fannon:
You mean if you configured it in .cdsrc but the link to the file does not exist?

Yes, would be good to point that out, also stating the name of the .cdsrc option and the value / path that cannot be found.

PR #93, opened 2024-11-12
feat: add cds logger

aramovic79:


lib/logger.js
```diff
@@ -0,0 +1,11 @@
+const cds = require("@sap/cds");
+
+const Logger = cds.log("ord-plugin", {
+    level: cds.env.DEBUG || process.env.DEBUG
```
aramovic79, line 4: @zongqichen : log level depends on DEBUG variable that can be set in two ways: as environment variable or in .npmrc.json document.
If it's value is truthy, the level is DEBUG, which means that all the messages with level DEBUG and higher will be shown in console; 
If it's value is not truthy, the default level is WARN, meaning that only logs with severity WARN and higher will be shown in the console.

aramovic79: @zongqichen : This way we instantiate a single instance of logger and set it's level and name(label) **centrally**.

aramovic79: @zongqichen : 
```
const logger1 = cds.log("some name");
const logger2 = cds.log("some name");

console.log(`The statement that logger1 is equal to logger2 is ${logger1 === logger2}.`); 
// result: "The statement that logger1 is equal to logger2 is true."
```

zongqichen: Thanks, I like the logger.js file. And I'm surprised that `logger1 === logger2`, need to refresh my knowledge.

aramovic79:
> Hi @aramovic79 , I agree to have only two levels, but the levels are the others. My opinion is level INFO and ERROR are important. When you debug or monitor logs in production env, INFO and ERROR can provide clear message to people who look into it. When I debug in dev env, the DEBUG level logger can't support me more than debug mode in vscode. Based on my experience, I always ignore warning message, since for me the program is either succeed or failed. So we need ERROR level message which can wake up admin at night shift when the program broke out.

This is the current setup:

In dev environment, the DEBUG level is set to some truthy value(in .cdsrc.json or by setting DEBUG env. variable), which results in showing all of the following logs:

-     logger.**DEBUG**('Will be shown');
-     logger.**INFO**('Will be shown');
-     logger.**WARN**('Will be shown');
-     logger.**ERROR**('Will be shown');

In production, the default level is WARNING, which means that only the following logs will be shown:
-     logger.**WARN**('Will be shown');
-     logger.**ERROR**('Will be shown');

@zongqichen : Is your proposal to set default level in production to be ERROR(so to "consider" only the `logger.ERROR('Will be shown);`)? If yes, I'm fine with that ...

lib/logger.js
aramovic79: > I'm surprised that `logger1 === logger2`, need to refresh my knowledge.
Yes, this is the link: https://pages.github.tools.sap/cap/docs/node.js/cds-log#logger-id
```text
The loggers constructed by cds.log() are cached internally, and the same instances are returned on subsequent
 invocations of cds.log() with the same id. This allows to use and share the same logger in different modules.
```

zongqichen:
> > Hi @aramovic79 , I agree to have only two levels, but the levels are the others. My opinion is level INFO and ERROR are important. When you debug or monitor logs in production env, INFO and ERROR can provide clear message to people who look into it. When I debug in dev env, the DEBUG level logger can't support me more than debug mode in vscode. Based on my experience, I always ignore warning message, since for me the program is either succeed or failed. So we need ERROR level message which can wake up admin at night shift when the program broke out.
> 
> This is the current setup:
> 
> In dev environment, the DEBUG level is set to some truthy value(in .cdsrc.json or by setting DEBUG env. variable), which results in showing all of the following logs:
> 
> * ```
>   logger.**DEBUG**('Will be shown');
>   ```
> * ```
>   logger.**INFO**('Will be shown');
>   ```
> * ```
>   logger.**WARN**('Will be shown');
>   ```
> * ```
>   logger.**ERROR**('Will be shown');
>   ```
> 
> In production, the default level is WARNING, which means that only the following logs will be shown:
> 
> * ```
>   logger.**WARN**('Will be shown');
>   ```
> * ```
>   logger.**ERROR**('Will be shown');
>   ```
> 
> @zongqichen : Is your proposal to set default level in production to be ERROR(so to "consider" only the `logger.ERROR('Will be shown);`)? If yes, I'm fine with that ...

My proposal is even simpler. We don't need to distinguish dev or production environment, we only use INFO and ERROR when it is really necessary. It will keep our code and console clean. And I visit other repos in cap-js, there are quite mix, some repo use debug and warn like [cds-typer](https://github.com/cap-js/cds-typer), some don't use at all like [change-tracking](https://github.com/cap-js/change-tracking). My preference is the second one. However, I don't strong opinion on it, what do you think @swennemers @Fannon ?

zongqichen:
Hi @aramovic79 , I forget to mention it, could you revisit all codes and replace all, console.error, console.info or new Error with logger? If you think it's out of the scope, you could raise an new issue for it. Thx

aramovic79:
> Hi @aramovic79 , I forget to mention it, could you revisit all codes and replace all, console.error, console.info or new Error with logger? If you think it's out of the scope, you could raise an new issue for it. Thx

Done as requested. Also, the issue #86 is fixed in this PR as well since it was partially related to logging. Could you please take a lot? Thank you!

lib/extendOrdWithCustom.js
```diff
@@ -49,21 +48,14 @@ function compareAndHandleCustomORDContentWithExistingContent(ordContent, customO
...
 function getCustomORDContent(appConfig) {
-    let customORDContent;
-
-    if (appConfig.env?.customOrdContentFile) {
-        customORDContent = require(path.join(cds.root, appConfig.env.customOrdContentFile));
-    }
-    return customORDContent;
+    if (!appConfig.env?.customOrdContentFile) return;
+    const pathToCustomORDContent = path.join(cds.root, appConfig.env?.customOrdContentFile);
+    if (fs.existsSync(pathToCustomORDContent))  return require(pathToCustomORDContent);
```
zongqichen: I would like have error log message here, including file not found and the path

zongqichen:
Hi @aramovic79 , thank you for replacing console log. There is still a small issue need to be adjust, https://cap.cloud.sap/docs/node.js/cds-log#recommendations. The construct format message in logger will be expensive for production env. The best practice is always use `,` to separate.

lib/extendOrdWithCustom.js
aramovic79: Done.

aramovic79:
> g. There is still a small issue need to be adjust, https://cap.cloud.sap/docs/node.js/cds-log#recommendations. The construct format message in logger will be expensive for production env. The best practice is always use `,` to separate.

Good point. Changed as proposed.

zongqichen (approved):
LGTM. Thx, it looks way better now

Issue #81, opened 2024-10-24
ORD and integration dependencies (consumed events)


RamIndia:
# cds import

- Typically, one imports events using `cds import <asyncAPI-file> [--as cds]`.
- This creates a model in `srv/external/<someModel>.cds`, which contains _all_ the Events, not just those of interest.
- Additionally, an entry is created in `cds.requires.<externalSrv>` that references this model.

Example:
Business Partner Events from S/4HANA: https://api.sap.com/event/CE_BUSINESSPARTNEREVENTS/overview

```bash
cds import ~/Downloads/CE_BUSINESSPARTNEREVENTS.json --as cds
```

Will create

```
├── srv
│   ├── external
│   │   ├── CE_BUSINESSPARTNEREVENTS.cds
│   │   └── CE_BUSINESSPARTNEREVENTS.json
```

and

```json
  "cds": {
    "requires": {
      "sap.s4.beh.businesspartner.v1.BusinessPartner": {
        "kind": "odata",
        "model": "srv/external/CE_BUSINESSPARTNEREVENTS"
      }
    }
  }
```

Problem: _All_ events are imported, not just the ones you're interested in.

# Solution

Create an own external service where you only include the events you're interested in.
Reference this model in your external service. For the example above:

```cds
using { sap.s4.beh.businesspartner.v1.BusinessPartner.Created.v1 as Created } from './CE_BUSINESSPARTNEREVENTS';
service OwnExtSrv {
  event BPCreated : projection on Created;
}
```

```json
  "cds": {
    "requires": {
      "sap.s4.beh.businesspartner.v1.BusinessPartner": {
        "kind": "odata",
        "model": "srv/external/OwnExtSrv"
      }
    }
  }
```

By checking the service definition of all external services (i.e. remote services in `cds.requires`), ORD can statically scan for all consumed events.

RoshniNaveenaS:
Update regarding the new approach:

Now the implementation is as follows:

We will look into all the services that are present in the `cds.requires`. Iterate over all of them and check for the CSN Models and detect if they have events declared and if it is present then add them to the corresponding integration dependencies.
@David-Kunz  please add if I missed anything in this.

Thanks and regards,
Roshni Naveena S

zongqichen:
Hi @Dipto-at-sap, could you update the status here? I saw you close the PR: https://github.com/cap-js/ord/pull/94#issuecomment-2876166683

Dipto-at-sap:
[...]

PR #94, opened 2024-11-13
[CDSADAPTERS-2187]: Added integration dependency for cds import for events

Dipto-at-sap:
JIRA: https://jira.tools.sap/browse/CDSADAPTERS-2187
Github issue: https://github.com/cap-js/ord/issues/81


lib/ord.js
```diff
@@ -165,6 +172,7 @@ module.exports = (csn) => {
     const eventResources = _getEventResources(linkedCsn, appConfig, packageIds);
     if (eventResources.length) {
         ordDocument.eventResources = eventResources;
+        ordDocument.integrationDependency = _getIntegrationDependencies(linkedCsn, appConfig, packageIds);
```
swennemers, line 175: this is not the right approach to build the integration dependencies, you have to read the package json/cds config for this, the information is not the same as the events

__tests__/__snapshots__/noOrdInCdsrc.test.js.snap
```diff
@@ -283,6 +283,7 @@ exports[`Tests for ORD document when .cdsrc.json has no \`ord\` property Success
       "title": "This is test Cinema Service title",
     },
   ],
+  "integrationDependency": [],
```
Fannon, line 286: This is actually named differently (plural). 
See https://pages.github.tools.sap/CentralEngineering/open-resource-discovery-specification/spec-v1/interfaces/document#ord-document_integrationdependencies

Please use the validator to check that the generated result is a valid ORD document

```suggestion
  "integrationDependencies": [],
```

lib/ord.js
```diff
@@ -30,8 +32,13 @@ const initializeAppConfig = (csn) => {
     const modelKeys = Object.keys(csn.definitions);
     const apiEndpoints = new Set();
     const events = [];
+    let externalServiceKeys = [];
+    if (packageJson.cds && packageJson.cds.requires) {
```
swennemers, line 36: What will happen, when the information is provided through the cdsrc.json and not via package json?

lib/templates.js
```diff
@@ -238,9 +238,48 @@ function _getPackageID(namespace, packageIds, resourceType) {
...
+    const ordExtensions = readORDExtensions(serviceDefinition);
+
+    const events = Object.keys(serviceDefinition.model.definitions)
+        .filter(event => {
+            const definition = serviceDefinition.model.definitions[event];
+            return definition["@topic"] && definition["@topic"].startsWith(serviceName) && !event.startsWith(serviceName);
+        })
+        .map(event => ({ eventType: serviceDefinition.model.definitions[event]["@topic"] }));
+
+    return {
```
swennemers, line 252: please follow the example of event and API templates to handle ordExtensions and don't explicitly access properties of ordExtensions and test that aspects can be overwritten as well (missing test)

lib/templates.js
```diff
@@ -238,9 +238,48 @@ function _getPackageID(namespace, packageIds, resourceType) {
...
+        ordId: `${appConfig.ordNamespace}:integrationDependency:${serviceName}:v1`,
+        title: ordExtensions.title ?? serviceDefinition['@title'] ?? "Integration Dependency sample title",
+        partOfPackage: _getPackageID(appConfig.ordNamespace, packageIds, "integration"),
+        version: ordExtensions.version ?? "1.0.0",
+        visibility: ordExtensions.visibility ?? "public",
+        lastUpdate: appConfig.lastUpdate,
+        description: ordExtensions.description ?? serviceDefinition['@description'] ?? "Integration Dependency sample description",
```
swennemers, line 259:
```suggestion
```
let's omit the description here

lib/templates.js
```diff
@@ -238,9 +238,48 @@ function _getPackageID(namespace, packageIds, resourceType) {
...
+        mandatory: ordExtensions.mandatory ?? true,
+        releaseStatus: ordExtensions.releaseStatus ?? "active",
+        aspects: [
+            {
+                title: ordExtensions.aspectTitle ?? "Integration Aspect sample title",
```
swennemers, line 264:
```suggestion
```
let's omit the title, it should come via ordExtension.

swennemers (changes requested):

__tests__/bookshop/srv/external/CE_BUSINESSPARTNEREVENTS.asyncapi.cds
```diff
@@ -0,0 +1,30 @@
...
+@cds.external : 'true'
+@topic : 'sap.s4.beh.businesspartner.v1.BusinessPartner.Changed.v1'
+event sap.s4.beh.businesspartner.v1.BusinessPartner.Changed.v1 {
+  BusinessPartner : LargeString;
+};
+
+@cds.external : 'true'
+@topic : 'sap.s4.beh.businesspartner.v1.BusinessPartner.Created.v1'
+event sap.s4.beh.businesspartner.v1.BusinessPartner.Created.v1 {
+  BusinessPartner : LargeString;
```
aramovic79, line 28: Maybe unrelated question(but still a bit confusing for me): Where is the `LargeString` coming from? How's this information derived from the content of the imported source file(`__tests__\bookshop\srv\external\CE_BUSINESSPARTNEREVENTS.asyncapi.json`)?

lib/ord.js
```diff
@@ -16,6 +17,7 @@ const _ = require("lodash");
 const cds = require("@sap/cds");
 const defaults = require("./defaults");
 const path = require("path");
+const { exists } = require('@sap/cds/lib/utils/cds-utils');
```
aramovic79, line 20: Minor finding: We use the `exists` as on line 25. No need for this line.

lib/ord.js
```diff
@@ -69,13 +76,26 @@ const initializeAppConfig = (csn) => {
         }
     }
 
+    if (packageJson.cds && packageJson.cds.requires) {
+        for (const key of externalServiceKeys) {
+            const service = packageJson.cds.requires[key];
+            if (
+                service?.kind?.startsWith("odata") &&
+                [".cds", ".csn"].some((ext) => exists(path.join(cds.root, `${service.model}${ext}`)))
```
aramovic79, line 84:
```suggestion
                [".cds", ".csn"].some((ext) => cds.utils.exists(path.join(cds.root, `${service.model}${ext}`)))
```

lib/templates.js
```diff
@@ -238,9 +238,48 @@ function _getPackageID(namespace, packageIds, resourceType) {
...
+        .filter(event => {
+            const definition = serviceDefinition.model.definitions[event];
+            return definition["@topic"] && definition["@topic"].startsWith(serviceName) && !event.startsWith(serviceName);
+        })
+        .map(event => ({ eventType: serviceDefinition.model.definitions[event]["@topic"] }));
+
+    return {
+        ordId: `${appConfig.ordNamespace}:integrationDependency:${serviceName}:v1`,
+        title: ordExtensions.title ?? serviceDefinition['@title'] ?? "Integration Dependency sample title",
+        partOfPackage: _getPackageID(appConfig.ordNamespace, packageIds, "integration"),
```
aramovic79, line 255: This will result in: "sap.sample:package:capireordsample-**api**:v1" which doesn't seem correct.

lib/templates.js
```diff
@@ -238,9 +238,48 @@ function _getPackageID(namespace, packageIds, resourceType) {
...
+        version: ordExtensions.version ?? "1.0.0",
+        visibility: ordExtensions.visibility ?? "public",
```
aramovic79, line 257:
```suggestion
        visibility: RESOURCE_VISIBILITY.public,
```

__tests__/bookshop/srv/external/CE_BUSINESSPARTNEREVENTS.asyncapi.cds
RoshniNaveenaS: Yes, since we are importing the event 
( and the property is LargeBinary) we have it that way

lib/templates.js
```diff
@@ -238,9 +238,48 @@ function _getPackageID(namespace, packageIds, resourceType) {
...
+    const events = Object.keys(serviceDefinition.model.definitions)
+        .filter(event => {
+            const definition = serviceDefinition.model.definitions[event];
+            return definition["@topic"] && definition["@topic"].startsWith(serviceName) && !event.startsWith(serviceName);
+        })
+        .map(event => ({ eventType: serviceDefinition.model.definitions[event]["@topic"] }));
+
+    return {
+        ordId: `${appConfig.ordNamespace}:integrationDependency:${serviceName}:v1`,
+        title: ordExtensions.title ?? serviceDefinition['@title'] ?? "Integration Dependency sample title",
```
zongqichen, line 254: In general, could you extend your test case with all `??` in this function? I only see the default value

lib/templates.js
```diff
@@ -238,9 +238,48 @@ function _getPackageID(namespace, packageIds, resourceType) {
...
+                title: ordExtensions.aspectTitle ?? "Integration Aspect sample title",
+                mandatory: ordExtensions.aspectMandatory ?? true,
+                eventResources: [
+                    {
+                        ordId: `${appConfig.ordNamespace}:eventResource:${serviceDefinition.name}:v1`,
+                        subset: events
+                    }
+                ]
+            }
+        ],
```
zongqichen, line 273: I also want to see the test case with aspects customized

lib/templates.js
Dipto-at-sap: In ORD_RESOURCE_TYPE, only an event and API type are defined, so can you tell me what the expected output is?

swennemers: For API Hub, the ORD packages must contain only resources from one ORD type. Therefore, we need an integration-dependency package.

lib/ord.js
Dipto-at-sap: if you have configurations in both package.json and cdsrc.json then it is considering both the configurations and both the objects are there in the integrationDependencies array
 
So, which one should be omitted or given more priority?
 
 Consider the following case:
Screenshot 2024-11-28 at 1 55 23 PM

David-Kunz:
I did a small example with

srv/external/myRemote.cds
```cds
service myRemote {
  event Foo { key ID: UUID };
}
```

package.json contains
```json
{
  "cds": {
    "requires": {
      "myRemote": {
        "kind": "odata",
        "model": "srv/external/myRemote.cds",
        "credentials": {
          "url": "http://www.foo.com"
        }
      }
  }
}
```

in custom code during bootstrapping:

```js
  const ext = await cds.connect.to('myRemote')
```

The problem is that the remote service definitions are not part of 

```js
  app.get("/open-resource-discovery/v1/documents/1", async (req, res) => {
      /* ... */
      const csn = await cds.load(cds.env.folders.srv);  // <-- doesn't include myRemote
 ```

Note: `cds.model` would include it, shouldn't this be used, also for performance reasons? In mtx scenarios, it should be the customer's model.

Since `csn` doesn't include the remote definitions (incl. events), `integrationDependencies` is empty.

File: lib/ord.js (hunk 3 above)
Comment (swennemers) [reply lib/ord.js:36]: @Dipto-at-sap ,  the comment was related to the issue @David-Kunz highlighted, you can have cds.requires in package.json and cdsrc.json the custom.ord.json is another aspect. 

With regards to the custom.ord.json, it must be consistent with the behavior described in https://github.com/cap-js/ord/blob/main/docs/ord.md. @zongqichen can support you in case of questions on that. Please reach out directly to him or @aramovic79 in his absence.

File: __tests__/bookshop/package.json [hunk 14]
```diff
@@ -15,5 +15,13 @@
...
   "scripts": {
     "start": "cds-serve"
+  },
+  "cds": {
+    "requires": {
+      "sap.s4.beh.businesspartner.v1.BusinessPartner": {
+        "kind": "odata",
+        "model": "srv/external/CE_BUSINESSPARTNEREVENTS.asyncapi"
+      }
+    }
```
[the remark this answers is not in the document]
swennemers: @David-Kunz , is this a correct cds.requires statement? Kind odata sounds like API required, but model is an async API, so events...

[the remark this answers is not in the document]
David-Kunz: Yes, that is weird. I've only seen `*.cds` or `*.json` files here.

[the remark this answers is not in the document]
David-Kunz: But I don't know if `*.asyncapi` is also supported...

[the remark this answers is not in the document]
Dipto-at-sap: it's indeed a json file, the name is given like that to state it's an asyncapi file.

lib/templates.js
Fannon: > @Fannon : Could you comment here pls? In this concrete case we are "importing/referencing" **event**(`sap.s4.beh.businesspartner.v1.BusinessPartner.Created.v1`), so it seems logical to me to be `event`(the second package ordId)...

Not sure I got the entire context, but an "event" like the example you gave are not something we have in ORD directly. An ORD Event Resource can describe several events (the ID you have is an event type). But in an ORD Integration Dependency, we can not just point to the Event Resource containing the events, we can also further restrict it, by stating which event types we need (we don't want to subscribe to all events of an entire event resource catalog).

__tests__/bookshop/package.json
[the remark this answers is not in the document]
swennemers: This statement doesn't match: https://pages.github.tools.sap/cap/docs/guides/messaging/s4#configure-cap 
Therefore I have my strong doubts the code is sitting on the right funcationality. @David-Kunz , can you help @Dipto-at-sap on how event consumption must be configured for a CAP app?

swennemers (changes requested):
I don't see my comments addressed.

Dipto-at-sap:
> I don't see my comments addressed.

@SebastianWolf-SAP for David's comment, currently we're assuming that the services will be given in package.json only and we'll set up a call for further discussion to address that. For now trying to get the other comments here resolved,.

lib/defaults.js
```diff
@@ -58,6 +58,7 @@ module.exports = {
             return [
                 createPackage(name, "-api:v1"),
                 createPackage(name, "-event:v1"),
+                createPackage(name,"-integration-dependency:v1")
```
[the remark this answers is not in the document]
Dipto-at-sap: @swennemers let me know which constant you're referring to. Here, I'm creating another package for integration dependency and defined the constant in the lib/constants.js just similar to what was there for event and api.

lib/ord.js
Dipto-at-sap: This is related to David's comment. So, we'll discuss it in the call.

lib/templates.js
Dipto-at-sap: In the case of createIntegrationDependencyTemplate, in the ordExtensions, there is no attribute related to event, so how will we get the eventType for all the consumed external events?

__tests__/bookshop/package.json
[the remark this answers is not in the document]
Dipto-at-sap: This was the requirement:
Screenshot 2024-12-03 at 2 41 17 PM

And for that issue, let me know if we're missing out something.

lib/defaults.js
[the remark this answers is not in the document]
swennemers: please reach out to @zongqichen and @aramovic79 to align

Dipto-at-sap:
Since it's an outdated PR, I am closing it for now.

PR #98, opened 2024-11-19
fix: apiResources may not include services with only events

aramovic79:


lib/ord.js
```diff
@@ -47,8 +48,11 @@ const initializeAppConfig = (csn) => {
                 break;
             // TODO: should be rewritten
             case ORD_RESOURCE_TYPE.entity:
-                if (!key.includes(".texts") && keyDefinition["@ODM.entityName"]) {
-                    odmEntity.push(createEntityTypeTemplate(keyDefinition));
+                if (!key.includes(".texts")) {
+                    entities.add(key);
+                    if (keyDefinition["@ODM.entityName"]) {
+                        odmEntities.push(createEntityTypeTemplate(keyDefinition));
+                    }
```
aramovic79, line 55: @Fannon @swennemers @RoshniNaveenaS @zongqichen : Does it make sense (in context of ORD doc creation) to have entities defined(and 'exposed' in service.cds file) without `@ODM.entityName` annotation?

aramovic79: > @aramovic79 - you mean whether they should be exposed in the CDS export? Or if they should be documented in the ORD Document as (local) Entity Types?

I mean - whether they should be considered when constructing `apiResources`.

lib/ord.js
```diff
@@ -95,7 +98,9 @@ const _getPackages = (policyLevel, appConfig) =>
             .slice(0, 1);
 
 const _getAPIResources = (csn, appConfig, packageIds) => {
+    if (appConfig.odmEntities.length === 0) return [];
```
swennemers, line 101: I doubt that this is the right filtering

lib/ord.js
```diff
@@ -95,7 +98,9 @@ const _getPackages = (policyLevel, appConfig) =>
...
     return appConfig.serviceNames
```
swennemers, line 102: I am missing a filtering based on a service containing only events and no entity or functions or action

aramovic79: > I am missing a filtering based on a service containing only events and no entity or functions or action

The logic is the following(pls check line 103): 
Some service is considered an `apiResource` _only_ if:
1. there is at least one entity existing, with **name starting with** that-service-name &&
2. the name of that entity **does not include** string `.texts`. 

Similarly, some service is considered an `eventResource` _only_ if:
1. there is at least one event existing, with **name starting with**  that-service-name

This is expressed by the lines from below. @swennemers @Fannon  : Do you agree?
`
            // TODO: should be rewritten
            case ORD_RESOURCE_TYPE.entity:
                if (!key.includes(".texts")) {
                    entities.add(key); // I've introduced the entities Set
                    if (keyDefinition["@ODM.entityName"]) {
                        odmEntities.push(createEntityTypeTemplate(keyDefinition));
                    }
                }
                break;
            case ORD_RESOURCE_TYPE.event:
                events.push(key);
                break;
`

`
.filter((serviceName) => appConfig.entities.some((entityName) => entityName.startsWith(serviceName)))
`

swennemers: I am missing functions + actions, that result in APIs. I don't get what this has to do with a variable called odmEntities or a  case ORD_RESOURCE_TYPE.entity.

lib/ord.js
```diff
@@ -40,32 +41,38 @@ const initializeAppConfig = (csn) => {
...
                 }
                 break;
             // TODO: should be rewritten
-            case ORD_RESOURCE_TYPE.entity:
-                if (!key.includes(".texts") && keyDefinition["@ODM.entityName"]) {
-                    odmEntity.push(createEntityTypeTemplate(keyDefinition));
+            case CDS_ELEMENT_KIND.entity:
+                if (!key.includes(".texts")) {
+                    endpoints.add(key);
+                    if (keyDefinition["@ODM.entityName"]) {
```
zongqichen, line 53: I would suggest that uses constant value ORD_EXTENSIONS_PREFIX defined in constans.js

lib/ord.js
```diff
@@ -40,32 +41,38 @@ const initializeAppConfig = (csn) => {
...
-
     return {
         env: cds.env["ord"],
         lastUpdate,
         appName,
+        endpoints: Array.from(endpoints),
         events,
         serviceNames,
-        odmEntity,
+        odmEntities: _.uniqBy(odmEntities, "ordId"),
```
zongqichen, line 75: same, we have constant value defined

lib/ord.js
```diff
@@ -24,10 +24,11 @@ const initializeAppConfig = (csn) => {
     const packageName = packageJson.name;
     const appName = packageJson.name.replace(/^[@]/, "").replace(/[@/]/g, "-");
     const modelKeys = Object.keys(csn.definitions);
+    const endpoints = new Set();
     const events = [];
     const serviceNames = [];
     const lastUpdate = getRFC3339Date();
-    let odmEntity = [];
+    let odmEntities = [];
```
zongqichen, line 31: I would like use const if odmEntities would not be assigned to other values

lib/ord.js
```diff
@@ -40,32 +41,38 @@ const initializeAppConfig = (csn) => {
...
         }
     }
 
-    odmEntity = _.uniqBy(odmEntity, "ordId");
-
     return {
         env: cds.env["ord"],
         lastUpdate,
         appName,
+        endpoints: Array.from(endpoints),
```
zongqichen, line 72: what is the meaning of endpoints? I would suggest provide more context in the name

__tests__/ordCdsrc.test.js
```diff
@@ -1,26 +1,19 @@
...
-});
-
-jest.mock("../lib/date", () => ({
-    getRFC3339Date: jest.fn(() => "2024-11-04T14:33:25+01:00")
-}));
-
-
 describe("Tests for default ORD document when .cdsrc.json is present", () => {
-    let csn;
+    let csn, ord;
```
zongqichen, line 5: what is the benefit of loading ord in beforeAll?

__tests__/unittest/extendOrdWithCustom.test.js
```diff
@@ -1,116 +1,151 @@
...
+                ordContent
+            );
             expect(result).toMatchSnapshot();
         });
 
-        it('should enhance the list of generated ord resources', () => {
-            const ordContent = { packages: [{ ordId: "sap.sm:package:smDataProducts:v1", localId: "smDataProductsV1" }] };
-            prepareTestEnvironment({}, appConfig, 'testCustomORDContentFileWithEnhanced.json');
-            const result = extendCustomORDContentIfExists(appConfig, ordContent);
+        it("should enhance the list of generated ord resources", () => {
```
zongqichen, line 73: I would suggest that set this format in .vscode settings, and then we don't need to take care of it in the future

swennemers (commented):
I am fine, when @zongqichen's comments are resolved

lib/ord.js
aramovic79: This constant/enum is CAP CDS/CSN specific, not directly related to ORD.
Here you can find an explanation why I've decided to give name CDS_ELEMENT_KIND: https://pages.github.tools.sap/cap/docs/cds/csn#definitions. 
There stands: definition `kind` can be: one of context, service, entity, type, action, function, or annotation
Maybe to rename into CSN_DEFINITION_KID or CDS_DEFINITION_KIND? Or you still prefer to rename `CDS_ELEMENT_KIND` into `ORD_EXTENSIONS_PREFIX`?

lib/ord.js
aramovic79: Done.

lib/ord.js
aramovic79: In the cds service files, entities, actions and functions will result into API endpoints. That's why I gave them the common name - "endpoints". On the other side, no endpoints => no apiResource.

__tests__/ordCdsrc.test.js
aramovic79: You can check by move up the import statement. Test will not work. It has to do with the why how the code is written. We first need to mock something and then to import it, if we want our tests to work.

__tests__/unittest/date.test.js
```diff
@@ -3,11 +3,16 @@ const { getRFC3339Date } = require('../../lib/date');
 const RFC3339_REGEX = /^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}(?:\.\d+)?(?:Z|[+-]\d{2}:\d{2})$/;
 
 describe('date', () => {
-    it('test getRFC3339Date', () => {
+    it('tests getRFC3339Date with offset', () => {
         const lastUpdate = getRFC3339Date();
         expect(lastUpdate).toMatch(RFC3339_REGEX);
     });
 
+    it('tests getRFC3339Date without offset', () => {
```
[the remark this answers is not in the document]
aramovic79: This is just to have all lines of code in date.js `covered`.

__tests__/unittest/extendOrdWithCustom.test.js
aramovic79: You mean - `formatter`, right? Sure, we can do that, but I don't think it's urgent to be done in this PR.

lib/ord.js
zongqichen: I mean write @ODM.entityName into constant value like ORD_EXTENSIONS_PREFIX.

lib/ord.js
zongqichen: I mean CONTENT_MERGE_KEY

lib/ord.js
zongqichen: thx, understand after your explanation, then could you rename it as apiEndpoints?

__tests__/unittest/extendOrdWithCustom.test.js
zongqichen: Following: https://github.com/cap-js/ord/issues/100

__tests__/unittest/date.test.js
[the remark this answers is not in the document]
zongqichen: Following: https://github.com/cap-js/ord/issues/99

Issue #88, opened 2024-10-30
Fix: cds command executed outside the root of application cause undefined error 


zongqichen:
I want to output ord document as static file and validate it in github action workflow:
```
    steps:
      - name: Set up Node.js
        uses: actions/setup-node@v4
        with:
          node-version: "18"

      - name: Install dependencies and prepare documents
        run: |
          npm install -g @sap/cds-dk
          npm install @sap/cds
          npm install https://github.com/cap-js/ord#fix_undefined
          cds compile ./ariba/srv --to ord -o ./tmp_documents/ariba.json
          cds compile ./cloud/srv --to ord -o ./tmp_documents/cloud.json
          cds compile ./erp/srv --to ord -o ./tmp_documents/erp.json
```
I create a debug branch and add two [logger info](https://github.com/cap-js/ord/blob/fix_undefined/lib/ord.js#L26), which prints out package name value.
The error shows:
```
packageName: undefined
[ord-plugin] - logger packageName undefined
TypeError: Cannot read properties of undefined (reading 'replace')
    at initializeAppConfig (/opt/actions-runner/_work/oid-cap-reference-app/oid-cap-reference-app/node_modules/@cap-js/ord/lib/ord.js:26:38)
    at module.exports (/opt/actions-runner/_work/oid-cap-reference-app/oid-cap-reference-app/node_modules/@cap-js/ord/lib/ord.js:151:23)
    at /opt/actions-runner/_work/_tool/node/18.20.3/x64/lib/node_modules/@sap/cds-dk/bin/compile/index.js:315:36
    at Generator.next (<anonymous>)
    at _write (/opt/actions-runner/_work/_tool/node/18.20.3/x64/lib/node_modules/@sap/cds-dk/lib/util/write.js:65:[28](https://github.tools.sap/CPA/oid-cap-reference-app/actions/runs/8966939/job/36012553?pr=49#step:6:29))
    at Object.to (/opt/actions-runner/_work/_tool/node/18.20.3/x64/lib/node_modules/@sap/cds-dk/lib/util/write.js:40:42)
    at /opt/actions-runner/_work/_tool/node/18.20.3/x64/lib/node_modules/@sap/cds-dk/lib/util/write.js:53:49
    at async Object.exec (/opt/actions-runner/_work/_tool/node/18.20.3/x64/lib/node_modules/@sap/cds-dk/bin/cds.js:91:16)
Error: Process completed with exit code 1.
```

## Diagnose:
[...]

PR #108, opened 2025-01-08
Fix: Correct packageJsonPath determination in ord.js

nirooxx:
This PR fixes the bug #88  in the `ord.js` where the `packageJsonPath` was incorrectly determined using `cds.root`, leading to errors when running the `cds compile` command in non-root contexts.
The issue occurred because `cds.root` does not always represent the working directory when `cds compile` is executed.
The fix updates the logic to use `process.cwd()` for determining the working directory, ensuring compatibility in various execution contexts.


aramovic79 (commented):
@nirooxx : I see 4 tests failing and the code coverage has decreased - now it's below 95%. 
cc: @zongqichen

lib/ord.js
```diff
@@ -19,13 +19,20 @@ const cds = require("@sap/cds");
 const defaults = require("./defaults");
 const path = require("path");
 
-const initializeAppConfig = (csn) => {
-    let packageJsonPath = path.join(cds.root, 'package.json')
+const initializeAppConfig = (csn, modelPath = process.cwd()) => {
+    if (typeof modelPath === 'object') {
```
aramovic79, line 23:
```suggestion
    if (typeof modelPath != 'string') { // typeof operator always returns string, so no need for checking type
```

lib/ord.js
```diff
@@ -19,13 +19,20 @@ const cds = require("@sap/cds");
...
+        modelPath = process.cwd(); 
+    }
+
+    if (typeof modelPath !== 'string') {
+        throw new TypeError(`Expected "modelPath" to be a string, but received ${typeof modelPath}: ${JSON.stringify(modelPath)}`);
+    }
```
aramovic79, line 29: @nirooxx : Please check the comment from above, it seems to me that these lines are needless.
cc: @zongqichen

lib/ord.js
```diff
@@ -161,9 +168,16 @@ function extractPackageIds(ordDocument) {
...
 
-module.exports = (csn) => {
+module.exports = (csn, modelPath = process.cwd()) => {
+    if (typeof modelPath === 'object') {
+        modelPath = process.cwd();
+    }
+
+    if (typeof modelPath !== 'string') {
+        throw new TypeError(`Expected "modelPath" to be a string, but received ${typeof modelPath}: ${JSON.stringify(modelPath)}`);
+    }
```
aramovic79, line 178: @nirooxx : Please check the comments from above.
cc: @zongqichen

lib/ord.js
```diff
@@ -161,9 +168,16 @@ function extractPackageIds(ordDocument) {
     return packageIds;
 }
 
-module.exports = (csn) => {
+module.exports = (csn, modelPath = process.cwd()) => {
```
aramovic79, line 171: @nirooxx : I think that the exported `ord` function is always called with the single parameter(`csn`), so the modelPath will always have default value (`process.cwd()`). In that sense, it seems to me needless to add `modelPath` as function parameter.
cc: @zongqichen

aramovic79 (changes requested):
@nirooxx : Reported some minor findings.
cc: @zongqichen @RoshniNaveenaS @Fannon

lib/ord.js
nirooxx: Done

lib/ord.js
nirooxx: Done

lib/ord.js
nirooxx: Done

__tests__/__snapshots__/mockedCsn.test.js.snap
```diff
@@ -21,11 +21,11 @@ exports[`Tests for ORD document generated out of mocked csn files Tests for ORD
         "supported": "yes",
       },
       "lastUpdate": "2024-11-04T14:33:25+01:00",
-      "ordId": "customer.capirebookshopordsample:apiResource:AdminService:v1",
+      "ordId": "customer.capjsord:apiResource:AdminService:v1",
```
zongqichen, line 24: Hi @nirooxx , in general, we should not change unittest case unless it is necessary. In this PR, we need to use package.json in bookshop instead of ord root package.json

lib/ord.js
```diff
@@ -19,8 +19,13 @@ const cds = require("@sap/cds");
 const defaults = require("./defaults");
 const path = require("path");
 
-const initializeAppConfig = (csn) => {
-    let packageJsonPath = path.join(cds.root, 'package.json')
+const initializeAppConfig = (csn, modelPath = process.cwd()) => {
```
zongqichen, line 22: I think we still need to keep cds.root, since we cds needs to know where is the root of package.json and other config files we want to load.

__tests__/__snapshots__/mockedCsn.test.js.snap
nirooxx: Hi @zongqichen , Good to know. The command `npm test -- -u` changed these files. I will rollback them.

zongqichen: Hi @nirooxx, all good, `-u` is to update test snapshot by current result. Only use npm test is sufficient.

lib/ord.js
nirooxx: I will update the implementation to use cds.root as the primary path, with process.cwd() serving as a fallback for scenarios where cds.root is unavailable, e.g. dynamic contexts or test environments.

aramovic79 (commented):
@nirooxx :
I have reproduced the issue #88 locally by running the command `cds compile ./erp/srv --to ord -o > tmp/erp_ord.json` **and** putting the break-point in node_modules/@cap-js/ord/lib/ord.js, line 17(function initializeAppConfig). 
In both cases, with using `cds.root` and with using `cds.root || process.cwd()` the obtained package.json corresponds to the one standing on the same path where the ongoing process is running. _In that sense, adding_ `process.cwd()` _as a fallback doesn't help._ 
Both functions/commands - `cds.root` and `process.cwd()` in this case are returning the same value - path on which the command `cds compile...` is executed.
Maybe @zongqichen you can give more context and details how you imagined to solve this issue?
To my understanding, the current behavior is correct and can stay as it was.
I don't think we should parse the command arguments (in this case it is path to the service/resource for which we want to generate an ord document: `./erp/srv`) and use them as a path to the package.json file, from which we then pick-up the application name(line: `const packageName = packageJson.name;`). I see no need for such over-engineering.

zongqichen:
Hi @nirooxx, I tested it locally, the issue still exists:
```
:~/work/oid-cap-reference-app$ npm install "github:cap-js/ord#fix/github-actions-ord"

added 84 packages, and audited 85 packages in 10s

15 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities
:~/work/oid-cap-reference-app$ cds compile ./ariba/srv/ --to ord -o .ariba.json
TypeError: Cannot read properties of undefined (reading 'replace')
    at initializeAppConfig (/work/oid-cap-reference-app/node_modules/@cap-js/ord/lib/ord.js:36:38)
    at module.exports (/work/oid-cap-reference-app/node_modules/@cap-js/ord/lib/ord.js:171:23)
    at .nvm/versions/node/v21.7.1/lib/node_modules/@sap/cds-dk/bin/compile/index.js:320:36
    at Generator.next (<anonymous>)
    at _write (/.nvm/versions/node/v21.7.1/lib/node_modules/@sap/cds-dk/lib/util/write.js:65:28)
    at Object.to /.nvm/versions/node/v21.7.1/lib/node_modules/@sap/cds-dk/lib/util/write.js:40:42)
    at /.nvm/versions/node/v21.7.1/lib/node_modules/@sap/cds-dk/lib/util/write.js:53:49
    at async Object.exec (/.nvm/versions/node/v21.7.1/lib/node_modules/@sap/cds-dk/bin/cds.js:94:16)
:~/work/oid-cap-reference-app$
```
my expectation is to set the cds.root with correct value, in this case is `/work/oid-cap-reference-app/ariba/`, instead of where I execute it. I agree with @aramovic79, if it is not worthwhile to spend too much effort on this issue, we can close it.

Issue #119, opened 2025-02-02
Fix: Don't create event resources when there are no events in CAP


zongqichen:
When models lack [`eventResources`](https://github.com/cap-js/ord/blob/main/lib/ord.js#L150) definitions, the ORD loader initializes empty packages due to overstrict conditional logic:

```javascript
// lib/ord.js (Line 150)
if (
  _getAPIResources(linkedCsn, appConfig).length &&  // AND condition
  _getEventResources(linkedCsn, appConfig).length   // ← Problematic check
) {
  ordDocument.packages = _getPackages(...);
}
```
It will cause following issues:
- packagesIds is empty
- no partOfPackage in entityTypes

We need to investigate why do we need to consider eventResouces when get packages and extend fully tests. At the moment, ord test does not cover no event testcase.

swennemers:
Please have 2 test cases: 
1. with only api resource
2. with only event resource

to ensure having any package calculation is independent of other resource types being available.

swennemers:
Only packages with at least one resource should be generated.

Please add test cases also for different visibilities (public + internal).

swennemers:
todo - retest issue

Issue #61, opened 2024-09-26
Feat: Support publishing to Business Accelerator Hub (api.sap.com)


Fannon:
For [Business Accelerator Hub](https://api.sap.com/) (BAH) publishing, we need to split the ORD Packages from one package (current behavior) into one package per ORD resource type. Most notably this means that we need a dedicated package for assigning APIs and another one for the events.

swennemers:
It should be there, also as default. But let's do a test run with our ORA sample apps once moved to the plugin.

Fannon:
> It should be there, also as default. But let's do a test run with our ORA sample apps once moved to the plugin.

Daniel Hutzel was explicitly against having this on by default.

swennemers:
well... the code looks differently

Fannon:
> well... the code looks differently

Ok, but to not run into troubles here, let's go with that. It could work like this:

* By default, just create one package. The name is derived from `package.json` name. 
* Add an option to support publishing to BAH. Not sure how to phrase it, should it be something like `separateOrdPackages`: `true` or more like `supportBusinessAcceleratorHub`: `true`?
* If this option is set, we need to create one Package per ORD resource type that has at least one instance. We can just clone the default package and suffix the ORD ID and title / descriptions with the content type. Like "... API resources".

Fannon:
@aramovic79 The rules which are most important are here: https://pages.github.tools.sap/CentralEngineering/open-resource-discovery-specification/spec-extensions/policy-levels/sap-core-v1#package 

If you validate with `sap:core:v1` some problems should already pointed out. I hope it's mostly about splitting up packages up correctly, according to the constraints. 

Example: You have one package, containing APIs, Events and Consumption Bundles.

Now you need to split it up into 3 Packages, containing only the same kind.
If you have mixed `visibility`, you again need to split up. So it could be 6 packages in the end. 

E.g. MyPackage -> MyPackageAPIs, MyPackageEvents, MyPackageConsumptionBundles

Now with visibility: MyPackageAPIsInternal,  MyPackageEventsInternal, MyPackageConsumptionBundlesInternal. 

I would just append "Internal" and treat "public" as the default.

PR #133, opened 2025-02-25
feat: Support publishing to Business Accelerator Hub (#61)

nirooxx:
Implements package separation and visibility handling for ORD compliance with Business Accelerator Hub.

- Splits APIs and Events into distinct packages
- Ensures visibility-based package assignment (`public` vs. `internal`)
- Implements dynamic package descriptions
- Improves package filtering to include only assigned resources


lib/defaults.js
```diff
@@ -23,6 +22,21 @@ const nameWithSpaces = (name) => {
...
 
+/**
+ * Generates dynamic description and short description for a package based on its type.
+ * @param {string} name - The package name.
+ * @param {string} type - The package type (API, Event, Consumption Bundle, etc.).
+ * @param {boolean} isInternal - Defines if the package is internal or public.
+ * @returns {object} Object containing the description and short description.
+ */
+const generatePackageDescriptions = (name, type, isInternal) => {
+    const visibility = isInternal ? "internal" : "public";
```
aramovic79, line 33: I think there is yet another [visibility](https://pages.github.tools.sap/CentralEngineering/open-resource-discovery-specification/spec-v1/interfaces/document#api-resource_visibility) to be considered.

lib/defaults.js
```diff
@@ -59,10 +73,16 @@ module.exports = {
...
+                createPackage(name, "-api:v1", "APIs"),
+                createPackage(name, "-api-internal:v1", "APIs", true),
+                createPackage(name, "-event:v1", "Events"),
+                createPackage(name, "-event-internal:v1", "Events", true),
+                createPackage(name, "-consumptionBundle:v1", "Consumption Bundles"),
+                createPackage(name, "-consumptionBundle-internal:v1", "Consumption Bundles", true),
+                createPackage(name, "-integrationDependency:v1", "Integration Dependencies"),
+                createPackage(name, "-integrationDependency-internal:v1", "Integration Dependencies", true),
+                createPackage(name, "-entityType:v1", "Entity Types"),
+                createPackage(name, "-entityType-internal:v1", "Entity Types", true),
```
aramovic79, line 85: @nirooxx: Not sure if it the right approach to "by default" generate packages for all 3 visibilities.
I think we should first check e.g. if there is at least one api/eventResource(or consumptionBundle..) with some visibility(public, internal, private) and then generate the corresponding package. I see no value of having an empty package, which will almost always happen if we leave the current implementation. Also, it doesn't make sense to me to have entityType with visibility `public` which is "referenced"(e.g. by entityTypeMappings[N].entityTypeTargets[N].ordId by some `private` apiResource. 
@Fannon @pavelkornev @swennemers : Do you agree?

swennemers: Please avoid empty packages.

For the reference, you should care about ghe visibility. The tool using ORD has to filter accordingly. It makes sense to me to have entityType with visibility public which is "referenced"(e.g. by entityTypeMappings[N].entityTypeTargets[N].ordId by some private apiResource.

nirooxx: @aramovic79 : At the end of the process, _filterUnusedPackages() in ord.js removes any unused packages to ensure that only those referenced in API, Event, or Consumption Resources are included in the ORD document.

lib/defaults.js
nirooxx: If private is also needed, I’ll adjust the logic accordingly.

lib/defaults.js
nirooxx: @aramovic79 @zongqichen
I have clarified the remaining open questions with Pavel:

- DataProducts should also be assigned to public/internal packages, similar to APIs and Events.

- If a public DataProduct references a private EntityType, the DataProduct itself must be private.

- Consumption Bundles remain unchanged, as they are not published to BAH.

- BAH treats an entire package as internal if it contains at least one internal/private artifact. 
   - To ensure visibility compliance, APIs, Events, EntityTypes, and DataProducts must be strictly separated into public and internal packages.

swennemers:
can you please check if you can resolve #119 in this PR?

nirooxx:
> can you please check if you can resolve #119 in this PR?

The condition has already been updated to use `||`, ensuring packages are generated even if only API or Event resources exist.

nirooxx:
> > > can you please check if you can resolve #119 in this PR?
> > 
> > 
> > The condition has already been updated to use `||`, ensuring packages are generated even if only API or Event resources exist.
> 
> Do you have test cases created for the issue?

The condition update is already covered. I’ve added a test to verify package generation without eventResources.

__tests__/mockedCsn.test.js
```diff
@@ -70,27 +74,28 @@ describe("Tests for ORD document generated out of mocked csn files", () => {
             const document = ord(csn);
 
             expect(document).not.toBeUndefined();
-            expect(document.entityTypes).toHaveLength(1);
+            expect(document.entityTypes).toHaveLength(2);
             expect(document.entityTypes[0].partOfPackage).toEqual(expect.stringContaining("entityType"));
             expect(document.entityTypes[0].level).toEqual(expect.stringContaining("root-entity"));
-            expect(document.apiResources[0].entityTypeMappings[0].entityTypeTargets).toEqual(expect.arrayContaining([
+
+       /*     expect(document.apiResources[0]?.entityTypeMappings[0]?.entityTypeTargets).toEqual(expect.arrayContaining([
```
swennemers, line 81: why do you have commented test code here? Please fix.

lib/templates.js
```diff
@@ -184,15 +184,59 @@ const createGroupsTemplateForService = (serviceName, serviceDefinition, appConfi
...
 const createEntityTypeTemplate = (appConfig, packageIds, entity) => {
     const ordExtensions = readORDExtensions(entity);
+    const visibility = ordExtensions.visibility || RESOURCE_VISIBILITY.public;
+
+    // Check whether the EntityType is referenced by a restrictive API
+    const referencedByRestrictedAPI = (Array.isArray(appConfig.apiResources) ? appConfig.apiResources : []).some(api =>
+        api.entityTypeMappings?.some(mapping =>
+            mapping.entityTypeTargets.some(target => target.ordId === entity.ordId)
+        ) && [RESOURCE_VISIBILITY.private, RESOURCE_VISIBILITY.internal].includes(api.visibility)
+    );
```
swennemers, line 203: Please generalize this check across APIs, Events and Data Products (you can pass a getEntity function to get the difference between resources with entityTypeMappings and plain entityType).

lib/templates.js
```diff
@@ -184,15 +184,59 @@ const createGroupsTemplateForService = (serviceName, serviceDefinition, appConfi
...
+    // Adopt the most restrictive visibility
+    let finalVisibility = visibility;
+    if (referencedByRestrictedAPI || referencedByRestrictedEvent || referencedByRestrictedDataProduct) {
+        if (referencedByRestrictedAPI && appConfig.apiResources.some(api => api.visibility === RESOURCE_VISIBILITY.private) ||
+            referencedByRestrictedEvent && appConfig.eventResources.some(event => event.visibility === RESOURCE_VISIBILITY.private) ||
+            referencedByRestrictedDataProduct && appConfig.dataProducts.some(dp => dp.visibility === RESOURCE_VISIBILITY.private)) {
+            finalVisibility = RESOURCE_VISIBILITY.private;
+        } else {
+            finalVisibility = RESOURCE_VISIBILITY.internal;
+        }
```
swennemers, line 227: The visibility logic should be the other way around. If it is referenced by any public resource, the entity is public. if it is referenced by no public, but any internal, it is internal, if it is reference by only private resources, it is private.

swennemers: and this logic can be combined with the functions above

swennemers: and I miss clear test cases that check this logic. they might exist, but I don't see them directly.

swennemers (changes requested):
sorry, some parts of the logic are wrong, 
The code and the tests are not clear.
Please fix

PR #136, opened 2025-03-21
Fix: Not able to consistently (or globally) overwrite vendor and product

zongqichen:


swennemers (commented):
please update the docu in https://github.com/cap-js/ord/blob/main/docs/ord.md.

PR #139, opened 2025-04-01
Fix: Not able to consistently (or globally) overwrite vendor and product

zongqichen:


swennemers:
please update the ORD configuration documentation as well otherwise it looks good

docs/ord.md
```diff
@@ -13,7 +13,18 @@ Example:
...
+    "policyLevel": "sap:core:v1",
+    "products": [
+        {
+            "ordId": "sap:product:eb.bm.tests:",
+            "vendor": "sap:vendor:SAP:"
+        }
+    ],
+    "packages": [
+        {
+            "vendor": "sap:vendor:SAP:"
```
swennemers, line 25: please add a statement, that the overwrites have an effect on products not being generated and packages to reference this product.

lib/ord.js
```diff
@@ -94,8 +94,14 @@ const _getPolicyLevel = (appConfig) =>
 const _getDescription = (appConfig) =>
     appConfig.env?.description || defaults.description;
 
-const _getProducts = (appConfig) =>
-    appConfig.env?.products || defaults.products(appConfig.packageName);
+const _getProducts = (appConfig) => {
```
swennemers, line 97: I think, we miss that we don't expose the product anymore, if the overwrite happened?!

swennemers (approved):

PR #140, opened 2025-04-10
Fix: display endpoint in index page

MariusPerleSAP:
- using CAP services to provide endpoints
- adding CAP services to package.json to be exposed to application
- removed duplication dependency in package.json


README.md
```diff
@@ -126,6 +126,17 @@ cds compile <path to srv folder> --to ord [-o] [destinationFilePath]
...
+Set environment variable `ORD_INDEX_HTML` to `true` to enable the ORD endpoints in the index.html file.
+You can set environment variable in the terminal, IDE config or in the `.env` file.
+
+Add a new ```ord-service.cds``` file to ```ord``` folder with the following content:
+```
+// This helps CAP to discover already existing endpoint for ORD
+@rest @path:'/.well-known/open-resource-discovery'
+service OrdService {}
+```
+
```
swennemers, line 139: this is a missconception in the requirement! The service is not to be a configuration option, but a replacement for the plain express endpoints

swennemers (dismissed):
@MariusPerleSAP please reach out, this goes into a strange direction

lib/ord-service.cds
```diff
@@ -0,0 +1,14 @@
...
+    entity documents {
+        key id : String;
+    }
+
+    function api() returns {};
+}
+
+@rest  @path: '/.well-known/open-resource-discovery'
+@ORD.Extensions.implementationStandard: 'sap:ord-document-api:v1'
+service OpenResourceDiscoveryService {}
```
MariusPerleSAP, line 14: it is also possible by breaking CAP conventions to get this endpoint with CAP service OrdService

lib/es-module.mjs
```diff
@@ -0,0 +1,7 @@
+// matches commonjs.js to es-module so that we can use the same code in cap services
+export { default as defaults } from "./defaults.js";
+export { default as getMetadata } from "./metaData.js";
+export { default as ord } from "./ord.js";
+export { default as plugin } from "./plugin.js";
+export { default as authentication } from "./authentication.js";
+export { default as logger } from "./logger.js";
```
swennemers, line 7: please only do this, if you plan to migrate now, step by step to mjs.

lib/ord-service.cds
swennemers: do it as you prefer.

swennemers:
@danjoa does this match the proposal how the plugin should be implemented as a service now?

lib/es-module.mjs
MariusPerleSAP: current CAP configuration of dev environment does not detect .cjs files.
Runtime currently treats .js files related to CAP services as es modules.

This matching is therefore required to connect custom code to our services.

My Recommendation is to enable a dual support.
Example how to do it: https://github.com/nodejs/package-examples

MariusPerleSAP:
Running ``` cds compile srv --to ord``` will not detect services provided by plugins.

This current approach does inject a CAP service. This results to a discrepancy between endpoint ORD document & CLI generated ORD document.

aramovic79 (approved):
Except 2 open questions/comments given by @MariusPerleSAP, this PR looks ok to me.

swennemers (approved):

zongqichen (approved):
LGTM

PR #141, opened 2025-04-23
Fix: additional endpoints for service documentation

MariusPerleSAP:
Additional PR to adress another imrpovement mentiod in issue 78.

This should be merged after #140 is completed. Base branch will update automatically after merge of #140 

Here we add additional endpoints to get 
- edmx
- openapi
- asyncapi
documents of services if these exist.


lib/ord-service.mjs
```diff
@@ -0,0 +1,127 @@
...
+        options: {
+            service,
+            as: 'str',
+            messages: []
+        },
+        csn
+    }
+}
+
+export class OrdService extends cds.ApplicationService {
```
aramovic79, line 22: This is what I meant when I said in today's meeting that we should split GET request handlers/routes: 
`
export class OrdService extends cds.ApplicationService {
    async init() {
        // COMMENT: We should be consistent with using try/catch here; so, we should use try/catch on all three places or nowhere
        cds.app.get(`${this.path}/documents/ord-document`, authentication.authenticate, async (_, res) => {
            const csn = cds.context?.model || cds.model;
            const data = ord(csn);
            return res.status(200).send(data);
        })


        /**
         * Middleware to handle the request for openapi, asyncapi, edmx documents
         */
        cds.app.get(`${this.path}/:ordId/:service`, authentication.authenticate, async (req, res) => {
            try {
                const { contentType, response } = await getMetadata(req.url);
                /* COMMENT: I still think there is no need for metaData.js file in this particular case
                 * I would even rewrite the getMetadata function - it can be much simpler since we already have some
                 * pre-filtering here
                 */
                return res.status(200).contentType(contentType).send(response);
            } catch (error) {
                logger.Logger.error(error, 'Error while processing the resource definition document');
                return res.status(500).send(error.message);
            }
        }
        );

        return super.init()
    }
}
`

MariusPerleSAP:
Closing this PR since its propose is no longer relevant.
Using feedback to improve #140

PR #145, opened 2025-04-29
Feat: compile resource definition files

zongqichen:


lib/metaData.js
```diff
@@ -4,11 +4,11 @@ const { compile: asyncapi } = require('@cap-js/asyncapi');
 const { COMPILER_TYPES } = require('./constants');
 const { Logger } = require('./logger');
 
-module.exports = async (data) => {
+module.exports = async (data, model = null) => {
     const parts = data?.split("/").pop().replace(/\.json$/, '').split(".");
     const compilerType = parts.pop();
```
swennemers, line 9: missing csn interop as compiler option @stewsk to give hint, how to trigger CAP compiler

cds-plugin.js
```diff
@@ -1,22 +1,69 @@
...
+        const ord = await ordPromise;
+        const getMetadata = await getMetadataPromise;
+        const ordDocument = ord(model);
+
+        const promises = []
+        promises.push(
+            this.write(ordDocument).to('ord-document.json')
+        )
+
+        for (const apiResource of ordDocument.apiResources) {
```
swennemers, line 47: do the same for event resources

cds-plugin.js
```diff
@@ -1,22 +1,69 @@
...
+        get: _lazyRegisterCompileTarget,
+        configurable: true,
+    });
 };
 
 registerORDCompileTarget();
+
+let ordPromise = import('./lib/es-module.mjs').then((esModule) => esModule.ord);
+let getMetadataPromise = import('./lib/es-module.mjs').then((esModule) => esModule.getMetadata);
+cds.build?.register?.('ord', class OrdBuildPlugin extends cds.build.Plugin {
```
swennemers, line 27: let's extract the class into an own module

swennemers:
Can we have same unit tests?

lib/metaData.js
```diff
@@ -29,6 +29,14 @@ module.exports = async (data) => {
                 throw error;
             }
             break;
+        case COMPILER_TYPES.csn:
+            try {
+                responseFile = await cds.compile(csn).to["csn"](options);
```
zongqichen, line 34: @swennemers, is it the correct method to compile csn?

swennemers: @stewsk has to answer for the correct method to compile csn interop, I assume this code is wrong especially as it is not yet fully exposed by the cds compiler...

zongqichen: @swennemers ,yes, need to wait for csn interop compiled support. Right now, I leave csn as instead.

swennemers (approved):
LGTM, let's move the correct CSN Interop compilation to a follow up PR.

PR #149, opened 2025-05-06
Fix: resources/event entityTypeMappings

MariusPerleSAP:
This PR addresses the following:
- Fixes issue with entityTypeMappings containing all entities for all resources or events
- Ensures only related entities are included in entityTypeMappings for all resources or events

**Changes**:
- added `_getEntityTypeMappings` to contain reused mapping logic to gather and structure all related entities
- adjusted multiple tests


lib/templates.js
```diff
@@ -344,41 +334,48 @@ const createEventResourceTemplate = (serviceName, serviceDefinition, appConfig,
...
 
     return obj.visibility === RESOURCE_VISIBILITY.public || obj.visibility === RESOURCE_VISIBILITY.internal ? [obj] : [];
 };
 
+/**
+ * This is a function to create entityTypeMappings array for service or event resource.
+ * @param serviceDefinition The service/event definition object.
+ * @returns {[{entityTypeTargets: *}]|*[]} An array of entityTypeMappings objects or an empty array.
+ * @private
+ */
```
zongqichen, line 349: I would suggest remove unused comments for now. We have no plan yet to generate document, and it's hard to maintain it if we change it later. Your function name and variables name are good enough to understand the logic.

lib/templates.js
```diff
@@ -344,41 +334,48 @@ const createEventResourceTemplate = (serviceName, serviceDefinition, appConfig,
...
+    const entities = Object.values(serviceDefinition.entities).map(createEntityTypeMappingsItemTemplate)
+    const entityTypeTargets = _.uniqBy(entities, CONTENT_MERGE_KEY).filter(entity => entity !== undefined).map(({ordId}) => ({
+        ordId
+    }))
+    if (entityTypeTargets.length > 0) {
+        return [{entityTypeTargets}]
+    } else {
+        return []
+    }
+}
```
zongqichen, line 360: I would suggest add one more test case with composition and association, e.g.:
```
entity Customers : managed {
  key ID         : String;
  addresses      : Composition of many Addresses on addresses.customer = $self;
  incidents      : Association to many Incidents on incidents.customer = $self;
}

entity Addresses : cuid, managed {
  customer       : Association to Customers;
}
entity Incidents : cuid, managed {
  customer       : Association to Customers;
}
```

__tests__/unittest/templates.test.js
```diff
@@ -316,4 +311,18 @@ describe('templates', () => {
...
+
+    describe('getEntityTypeMappings', () => {
+        it('should clean up duplicates', () => {
+            const serviceDefinition = {
+                entities: [{}, {}, {}]
+            }
+            serviceDefinition.entities[0][ORD_ODM_ENTITY_NAME_ANNOTATION] = 'Something';
+            serviceDefinition.entities[1][ENTITY_RELATIONSHIP_ANNOTATION] = 'sap.sm:entityType:Else:v1';
+            serviceDefinition.entities[2][ENTITY_RELATIONSHIP_ANNOTATION] = 'sap.odm:Something:v1';
+            expect(_getEntityTypeMappings(serviceDefinition)).toEqual([{
```
zongqichen, line 323: We can use `toMatchSnapshot()` as same as other test

__tests__/__snapshots__/mockedCsn.test.js.snap
```diff
@@ -57,15 +57,7 @@ exports[`Tests for ORD document generated out of mocked csn files Tests for ORD
...
-      "entityTypeMappings": [
-        {
-          "entityTypeTargets": [
-            {
-              "ordId": "sap.odm:entityType:odm.bookshop.Authors:v1",
-            },
-          ],
-        },
-      ],
+      "entityTypeMappings": [],
```
zongqichen, line 60: @swennemers, if entityTypeMappings is emtpy, shall we remove the property? I think right now, our plugin follows lazy mode, which always generates the property if it needed.

zongqichen (commented):
In general LGTM, just few comments.

lib/templates.js
MariusPerleSAP: good catch

MariusPerleSAP: enhanced now in [7cf041f](https://github.com/cap-js/ord/pull/149/commits/7cf041fccd75e8c7085a0c46d9e38bb4ffd8da39)

lib/templates.js
```diff
@@ -344,41 +334,62 @@ const createEventResourceTemplate = (serviceName, serviceDefinition, appConfig,
...
+        if (processedEntities.includes(target)) {
+            return;
+        }
+        processedEntities = [...processedEntities, target]
+        assertionsTodo.push(entity)
+    })
+
+    return [currentEntity, ...assertionsTodo.flatMap(entity => _assertionsFlatList(entity, processedEntities))]
+}
+
```
zongqichen, line 375:
```suggestion
function _flattenEntityGraph(currentEntity, processedEntities = []) {
    if(!currentEntity.associations) {
        return [currentEntity]
    }
    const entityAssociationTargets = Object.values(currentEntity.associations).map(association => ({target: association.target, entity: association._target}))

    const entityToProcess = []
    entityAssociationTargets.forEach(({target, entity}) => {
        if (processedEntities.includes(target)) {
            return;
        }
        processedEntities = [...processedEntities, target]
        entityToProcess.push(entity)
    })

    return [currentEntity, ...entityToProcess.flatMap(entity => _flattenEntityGraph(entity, processedEntities))]
}

```
Implementation is good. I would suggest that names the variables more clear.

lib/templates.js
```diff
@@ -344,41 +334,62 @@ const createEventResourceTemplate = (serviceName, serviceDefinition, appConfig,
...
+    }))
+    if (entityTypeTargets.length > 0) {
+        return [{entityTypeTargets}]
+    } else {
+        return []
+    }
+}
+
+function _assertionsFlatList(currentEntity, processedEntities = []) {
+    if(!currentEntity.associations) {
```
zongqichen, line 359: I'm thinking what about compositions? Also needs sufficient testcase.

__tests__/unittest/templates.test.js
```diff
@@ -315,5 +310,58 @@ describe('templates', () => {
             expect(eventResourceTemplate).toMatchSnapshot();
             expect(eventResourceTemplate).toEqual([]);
         });
+
+        it('should find composition and assertion entities for related service', () => {
```
zongqichen, line 314:
```suggestion
        it('should find composition and association entities for related service', () => {
```

__tests__/unittest/templates.test.js
```diff
@@ -315,5 +310,58 @@ describe('templates', () => {
...
+            const serviceName = 'MyService';
+            linkedModel = cds.linked(`
```
zongqichen, line 316: @swennemers , how to design the model to be reflected the following structure, with two more entityTypeTargets objects?:
```
      "entityTypeMappings": [
        {
          "entityTypeTargets": [
            {
              "ordId": "sap.odm:entityType:RetailTransaction:v1"
            }
          ]
        },
        {
          "entityTypeTargets": [
            {
              "correlationId": "sap.csnexposure:entity:RetailTransaction"
            }
          ]
        }
      ],
```

zongqichen (commented):
Few more discusstions

lib/templates.js
```diff
@@ -344,41 +334,62 @@ const createEventResourceTemplate = (serviceName, serviceDefinition, appConfig,
...
+    const assertionsTodo = []
+    entityAssociationTargets.forEach(({target, entity}) => {
+        if (processedEntities.includes(target)) {
+            return;
+        }
+        processedEntities = [...processedEntities, target]
+        assertionsTodo.push(entity)
+    })
+
+    return [currentEntity, ...assertionsTodo.flatMap(entity => _assertionsFlatList(entity, processedEntities))]
```
zongqichen: @swennemers, how does entityTypeTargets look like to this case?
```
service A {
    entity b as projection on B;
}

entity B {
    c: Association to C;
}

entity C {
    d: Composition to D;
}

entity D {
    key ID: Integer;
}
```

__tests__/__snapshots__/mockedCsn.test.js.snap
swennemers: please check with the validator

__tests__/unittest/templates.test.js
swennemers: We don't support the correlation id use case in the plugin. 
I quickly discussed with @MariusPerleSAP , having multiple targets only makes sense, when you have different apiModelSelectors. As we don't have them in the model, we basically have the "empty" apiModelSelector that means somewhere in the whole api you find the targets, so it is fine to have multiple objects in the same entityTypeTargets. @Fannon , please confirm if that is also your understanding of the ORD spec here.

lib/templates.js
swennemers: without annotations (ODM or ER) no entity type targets

lib/templates.js
```diff
@@ -344,41 +334,62 @@ const createEventResourceTemplate = (serviceName, serviceDefinition, appConfig,
...
     if (!packageIds) return;
 
     if (resourceType) {
         return packageIds.find(id => {
             if (visibility === RESOURCE_VISIBILITY.public) {
                 return id.includes(resourceType) &&
-                       !id.includes("-internal") &&
-                       !id.includes("-private");
+                    !id.includes("-internal") &&
+                    !id.includes("-private");
```
swennemers, line 384: @MariusPerleSAP and @zongqichen , please align on whitespace/prettier settings.

__tests__/__snapshots__/mockedCsn.test.js.snap
zongqichen: Validator doesn't complain it. But we have no special reason to only have this optional property with empty value

__tests__/unittest/templates.test.js
```diff
@@ -315,5 +310,140 @@ describe('templates', () => {
...
+            const packageIds = ['customer.testNamespace:package:test:v1'];
+            const apiResourceTemplate = createAPIResourceTemplate(serviceName, srvDefinition, appConfig, packageIds);
+
+            expect(apiResourceTemplate).toMatchSnapshot();
+        });
+
+        it('should find composition on nested entities for related service', () => {
+            const serviceName = 'MyService';
+            linkedModel = cds.linked(`
+                entity SecureApps {
```
swennemers, line 399: what about adding the ODM + ER annotation also to this associated entity? => it should then appear in the entity type targets of the service as the service = API/Event can have entities that are compositions and associations of each other.

__tests__/unittest/templates.test.js
```diff
@@ -315,5 +310,140 @@ describe('templates', () => {
...
+                    key ID          : String;
+                    components      : Composition of many Components on components.app = $self;
+                }
+
+                @ODM.entityName: 'directCompositionOdmEntity'
```
swennemers, line 404:
```suggestion
                @ODM.entityName: 'DirectCompositionOdmEntity'
```
FYI: Naming convention is UpperCamelCase...

MariusPerleSAP: will adjust in all tests

lib/templates.js
MariusPerleSAP: As soon as Service A and entity D is annotated it will OrdID of D will display

zongqichen:
```
service A {
    entity b as projection on B;
}

@ODM.entityName: 'B'
entity B {
    c: Association to C;
}

entity C {
    d: Composition to D;
}

@ODM.entityName: 'D'
entity D {
    key ID: Integer;
}
```

entityTypeTargets:  B and D

__tests__/unittest/templates.test.js
zongqichen: Testcase: A -> B -> C -> A

MariusPerleSAP (commented):
Test "Successfully create ORD Documents with defaults" is affected by other tests.

swennemers (approved):

PR #150, opened 2025-05-06
call forEffective

stewsk:


zongqichen:
Hi @stewsk,
I test it local with few lines patch, I think the file is still not csn-interop right? I didn't find the property `csnInteropEffective`. 
```
{
  "contentType": "application/json",
  "response": {
    "definitions": {
    ...
    },
    "meta": {
      "creator": "CDS Compiler v5.9.2",
      "flavor": "inferred"
    },
    "$version": "2.0"
  }
}
```

You can reproduce it by:
```
cd xmpl/
cds build --for ord
```

stewsk:
Hi @zongqichen , you are right, what we produce is not 100% CSN interop, but it is the closest that we have right now.
My proposal is to add the missing attributes directly here in the ORD plugin as a temporary workaround until the compiler can do it.
I can add the respective code to this PR if you agree.

Fannon:
> Hi @zongqichen , you are right, what we produce is not 100% CSN interop, but it is the closest that we have right now. My proposal is to add the missing attributes directly here in the ORD plugin as a temporary workaround until the compiler can do it. I can add the respective code to this PR if you agree.

Just adding the missing header attributes might not be enough, you'll have to look at all annotations and then rename some into the CSN Interop agreed annotation. E.g. `@title`, `@Common.label` have to be renamed to `@EndUserText.label` (https://sap.github.io/csn-interop-specification/spec-v1/extensions/end-user-text#endusertextlabel)

swennemers:
> Hi @zongqichen , you are right, what we produce is not 100% CSN interop, but it is the closest that we have right now. My proposal is to add the missing attributes directly here in the ORD plugin as a temporary workaround until the compiler can do it. I can add the respective code to this PR if you agree.

@stewsk , please add the patching either to this PR, or a follow up.

@zongqichen will help with the test case: fix the mocked test and provide e2e testing option in the xmpl app (by using the data product annotation on a service)

swennemers (approved):
ok, let's merge it to get the improvement in and do the formatting PR #151 done.

We still need to correct the csn interop. @stewsk please do that in a follow up PR.

zongqichen:
Hi @stewsk , I fix the test and add dataProduct annotation, you can also test locally.

When I run `cds build --for ord`, everything works fine, and the build completes successfully. However, when I run `cds w`, I encounter the following error:
```
CDS compilation failed (@sap/cds-compiler v5.9.2) ../../../incidents-app/srv/services.cds:7:10: Error: Element can't be virtual and an association or composition (in entity:“ProcessorService.Incidents”/element:“DraftAdministrativeData”)
```
The different behaver is this [line](https://github.com/cap-js/ord/blob/use-eff-csn/lib/metaData.js#L12):
```
const csn = this.model() || cds.services[serviceName]?.model;
```
`cds.services[serviceName]?.model` provides more keyword like: `DraftAdministrativeData`.  Is there a way to ensure consistent behavior between the two ways of getting model? Thx

xmpl/ord/custom.ord.json
```diff
@@ -67,5 +76,28 @@
...
+            "type": "primary",
+            "category": "business-object",
+            "outputPorts": [
+                {
+                    "ordId": "sap.sample:apiResource:manualImplementedService:v1"
+                }
+            ],
+            "responsible": "customer:group:capire-team",
+            "entityTypes": ["sap.odm:entityType:BusinessPartner:v1"]
+        }
```
swennemers, line 101: why did you add this? it is not necessary for the csn interop and data product production story as BDC will create the data product ORD

zongqichen: Test purpose, forgot to clean it...

PR #152, opened 2025-05-08
Feat: backward compatible (policyLevels to be plural)

zongqichen:
ORD root property `policyLevel` is [deprecated](https://pages.github.tools.sap/CentralEngineering/open-resource-discovery-specification/spec-v1/interfaces/document#ord-document_policylevel). We need to update it.

However, we need to be compatible with older versions.  Here is open discussion,

### DIFF
--- a/lib/defaults.js
+++ b/lib/defaults.js
@@ -45,13 +45,22 @@ const generateUniquePackageOrdId = (ordNamespace, name, tag, visibility) => {
 };
 
 /**
+ * Checks if at least one policy level in the array is SAP.
+ *
+ * @param {string[]} policyLevels - Array of policy levels.
+ */
+function hasSAPPolicyLevel(policyLevels) {
+    return policyLevels.some((policyLevel)=> policyLevel.split(":")[0].toLowerCase() === "sap");
+}
+
+/**
  * Module containing default configuration for ORD Document.
  * @module defaults
  */
 module.exports = {
     $schema: "https://open-resource-discovery.github.io/specification/spec-v1/interfaces/Document.schema.json",
     openResourceDiscovery: OPEN_RESOURCE_DISCOVERY_VERSION,
-    policyLevels: "none",
+    policyLevels: ["none"],
     description: "this is an application description",
     products: (name) => [
         {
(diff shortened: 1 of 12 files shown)


lib/defaults.js
```diff
@@ -51,7 +51,7 @@ const generateUniquePackageOrdId = (ordNamespace, name, tag, visibility) => {
 module.exports = {
     $schema: "https://open-resource-discovery.github.io/specification/spec-v1/interfaces/Document.schema.json",
     openResourceDiscovery: OPEN_RESOURCE_DISCOVERY_VERSION,
-    policyLevel: "none",
+    policyLevels: "none",
```
swennemers, line 54:
```suggestion
    policyLevels: ["none"],
```
See https://pages.github.tools.sap/CentralEngineering/open-resource-discovery-specification/spec-v1/interfaces/document#ord-document_policylevels, it is an array now

lib/defaults.js
```diff
@@ -62,12 +62,12 @@ module.exports = {
...
     ],
     groupTypeId: "sap.cds:service",
-    packages: function getPackageData(appConfig, policyLevel) {
+    packages: function getPackageData(appConfig, policyLevels) {
         const name = appConfig.appName;
         const ordNamespace = appConfig.ordNamespace;
         const productsOrdId = appConfig.existingProductORDId || appConfig.products?.[0]?.ordId;
         const vendor = appConfig.env?.packages?.[0]?.vendor;
-        if (policyLevel.split(":")[0].toLowerCase() === "sap") {
+        if (policyLevels.split(":")[0].toLowerCase() === "sap") {
```
swennemers, line 70: it is an array now!

swennemers:
@Fannon are we ready to remove policyLevel (singular)? Is the plural attribute supported by UCL?

Fannon:
> @Fannon are we ready to remove policyLevel (singular)? Is the plural attribute supported by UCL?

@swennemers yes, UCL will ignore the policyLevel(s) in the ORD document anyway, they always validate with `sap:base:v1`. Confirmed with Emma Karova.

swennemers (approved):
@zongqichen , please check and review from your end

zongqichen:
LGTM

PR #153, opened 2025-05-09
Fix: handle eventResources and apiResources

zongqichen:
To fix:
- no event -> no eventResources
- no service -> no apiResources
- if service set to be private -> no apiResources and no eventResources inside the service
- generate packages only if there are apiResources or eventResources
- only event in service, no apiResource and no group


swennemers:
> To fix:
> 
> * no event -> no eventResources
> * no service -> no apiResources
> * if service set to be private -> no apiResources and no eventResources inside the service
> * generate packages only if there are apiResources or eventResources
> * only event in service, no apiResource and no group

 only event in service, no apiResource but still a group

lib/templates.js
```diff
@@ -218,20 +164,21 @@ const createEntityTypeTemplate = (appConfig, packageIds, entity) => {
...
-/**
- * This is a template function to create API Resource object for API Resource Array.
- * Properties of an API resource can be overwritten by the ORD extensions. Example: visibility.
- * Ensures proper visibility compliance by checking associated EntityTypes.
-
- * @param {string} serviceName The name of the service.
- * @param {object} serviceDefinition The definition of the service
- * @param {object} appConfig - The application configuration.
- * @param {Array} packageIds - The available package identifiers.
- * @returns {Array} An array of objects for the API Resources.
```
swennemers, line 230: did you remove the jsdoc on purpose? the @param helps vs code to give better code completion...

zongqichen: Yes, I removed it intentionally. From my perspective, the jsdoc comments didn’t add much value and were more of a distraction. If the variable names aren’t self-explanatory, I believe we should focus on improving the naming instead. In general, I prefer to keep the code clean and simple rather than relying on verbose comments.

__tests__/__snapshots__/ord.e2e.test.js.snap
```diff
@@ -155,20 +155,42 @@ exports[`End-to-end test for ORD document Tests for default ORD document when .c
...
+      ],
+      "shortDescription": "AdminService.BookCreated event resource",
+      "title": "ODM capirebookshopordsample Events",
+      "version": "1.0.0",
+      "visibility": "public",
+    },
+    {
+      "description": "CAP Event resource describing events / messages.",
       "extensible": {
-        "supported": "yes",
```
swennemers, line 166: is this intended?

__tests__/__snapshots__/ord.e2e.test.js.snap
```diff
@@ -155,20 +155,42 @@ exports[`End-to-end test for ORD document Tests for default ORD document when .c
   "eventResources": [
     {
       "description": "CAP Event resource describing events / messages.",
-      "entityTypeMappings": {
-        "entityTypeTargets": [
-          {
-            "ordId": "sap.odm:entityType:test-from-extension:v1",
-          },
-        ],
```
swennemers, line 163: events can also have entityTypeMappings, does that still work?

lib/ord.js
```diff
@@ -22,78 +23,143 @@ const defaults = require("./defaults");
...
+    if (_shouldSkipIfServiceOnlyContainsEvents(serviceDefinition) || serviceDefinition["@cds.external"]) {
+        return null;
+    }
+    return apiResourceName;
+}
 
-const _getPolicyLevels = (
-    appConfig) => appConfig.env?.policyLevels ||
-    appConfig.env?.policyLevel && [appConfig.env?.policyLevel] ||
-    defaults.policyLevels;
```
swennemers, line 96: is this intended? I had introduced the fallback handling here, if someone still uses policyLevel singular.

__tests__/__snapshots__/ord.e2e.test.js.snap
zongqichen: From my understanding, for example:
```
service AdminService @(requires: 'authenticated-user') {
    entity Books   as projection on my.Books;
    entity Authors as projection on my.Authors;

    event BookCreated : {
        ID    : Integer;
        title : String @title: 'Title';
    };
}
```
The event BookCreated doesn't contain entity inside. Or it should be targeted as the entities of AdminService, like Books and Authors?

lib/templates.js
MariusPerleSAP: We need to investigate how to communicate the return type of a function. I recommend to switching to typescript if possible.

lib/templates.js
```diff
@@ -406,6 +343,23 @@ function _getPackageID(namespace, packageIds, resourceType, visibility = RESOURC
...
 }
 
+function _propagateORDVisibility(model) {
+    for (const [name, def] of Object.entries(model.definitions)) {
+        if (def.kind === CDS_ELEMENT_KIND.service && def[ORD_EXTENSIONS_PREFIX + "visibility"]) {
+            const visibility = def[ORD_EXTENSIONS_PREFIX + "visibility"];
+            for (const def of model.definitions) {
+                if (def.name.startsWith(name + ".")) {
+                    if (!def[ORD_EXTENSIONS_PREFIX + "visibility"]) {
+                        def[ORD_EXTENSIONS_PREFIX + "visibility"] = visibility;
```
MariusPerleSAP, line 353: Does this line to anything since:
`const visibility = def[ORD_EXTENSIONS_PREFIX + "visibility"];`leads to
`
def[ORD_EXTENSIONS_PREFIX + "visibility"] =  def[ORD_EXTENSIONS_PREFIX + "visibility"]
`

?

__tests__/__snapshots__/mockedCsn.test.js.snap
```diff
@@ -173,20 +173,42 @@ exports[`Tests for ORD document generated out of mocked csn files Tests for ORD
...
       "extensible": {
-        "supported": "yes",
+        "supported": "no",
       },
       "lastUpdate": "2024-11-04T14:33:25+01:00",
-      "ordId": "customer.capirebookshopordsample:eventResource:AdminService:v1",
+      "ordId": "customer.capirebookshopordsample:eventResource:AdminService.BookDeleted:v1",
       "partOfGroups": [
-        "sap.cds:service:customer.capirebookshopordsample:AdminService",
+        "sap.cds:service:customer.capirebookshopordsample:AdminService.BookDeleted",
```
MariusPerleSAP, line 211: There is a huge difference between snapshots. Here is one example that doesn't look right,

lib/ord.js
swennemers: well, there is a difference between the ORD compatibility and our plugins config compatible. therefore, removing it has the risk of disturbing a first adopter

lib/templates.js
zongqichen: I agree with it. However, we had discussion before about javascript or typescript. We decided to continue with javascript since CAP backend supports javascript, we have no clue how well is typescript supported. It's time to revisit and factor it.

lib/templates.js
zongqichen: Good point, the name and logic are not clear and simple, need to be refactored. But the logic is correct, here is the example of definitions:
```
AdminService: {
@ORD.Extensions.visibility: "private",
kind: "service",
},
AdminService.ServiceEvent: {
kind: "event",
},
OtherService: {
kind: "service"
}
...
``` 
If we find that AdminService is service and has annotation visibility, then we revisit the definitions and propagate this annotation to child-properties, if the name matches the pattern "AdminService." and it doesn't have visibility annotation.

__tests__/splitUpPackagesCsn.test.js
```diff
@@ -1,47 +0,0 @@
...
-    });
-
-    describe("Tests for ORD document for Business Accelerator Hub", () => {
-        test("Successfully create ORD Documents for Business Accelerator Hub", () => {
-            const document = ord(bahCsn);
-            expect(document).not.toBeUndefined();
-            document.apiResources.forEach((api) => delete api.lastUpdate);
-            document.eventResources.forEach((event) => delete event.lastUpdate);
-            document.consumptionBundles.forEach((con) => delete con.lastUpdate);
-            expect(document).toMatchSnapshot();
```
swennemers, line 39: Please confirm that we have tests for the package splitting (internal + public + ...) @zongqichen

lib/templates.js
swennemers: typing discussion to move into #154

__tests__/splitUpPackagesCsn.test.js
zongqichen: There are tests for package splitting: https://github.com/cap-js/ord/blob/main/__tests__/unittest/ordVisibilitySplit.test.js, however, I still restore splitUpPackagesCsn.test.js for safe.

swennemers (approved):

PR #167, opened 2025-06-05
Feat: use relative path for resource definitions URL and compatible with windows

zongqichen:


__tests__/unittest/build.test.js
```diff
@@ -152,4 +153,41 @@ describe("Build", () => {
...
+        };
+        const updatedOrdDocument = buildClass.postProcess(ordDocument);
+        expect(updatedOrdDocument.apiResources[0].resourceDefinitions[0].url).toBe(
+            path.join(
+                BUILD_DEFAULT_PATH,
+                "resource1_v1"
+            )
+        );
+        expect(updatedOrdDocument.eventResources[0].resourceDefinitions[0].url).toBe(
+            path.join(
```
swennemers, line 187: Can you please hard code the result to avoid mistakes in the result calculation of the test?

__tests__/unittest/build.test.js
```diff
@@ -110,7 +111,7 @@ describe("Build", () => {
     });
 
     it("should skip when OpenResourceDiscoveryService founded", async () => {
-        jest.spyOn(console, "log").mockImplementation(() => {});
+        jest.spyOn(console, "log").mockImplementation(() => { });
```
swennemers, line 114: Didn't you align on formatting?

zongqichen: Good catch, in the past, I want to avoid settings conflict with calesi project, so I set disable the settings in ord .vscode, now I added it back.

lib/build.js
```diff
@@ -38,12 +37,40 @@ module.exports = class OrdBuildPlugin extends cds_dk.build.Plugin {
...
+        _updateResourceUrls(clonedOrdDocument.eventResources);
+
+        return clonedOrdDocument;
+    }
+
+    _createRelativePath(url) {
+        let relative = url.split("/ord/v1").pop();
+        if (relative.startsWith("/")) relative = relative.slice(1);
+        relative = relative.replace(/:/g, "_");
+        return path.join(BUILD_DEFAULT_PATH, ...relative.split("/"));
```
swennemers, line 64: this doesn't look like a relative path

zongqichen: I see how does other team use this document, the prefix `gen/ord/` is not needed, if they copy and paste to other project.

swennemers (approved):

PR #171, opened 2025-06-18
Fix: set cds version to 8

zongqichen:


lucasborin:
Could you bump the minor release, please? This way, I can force npm to pull the new version.

Issue #176, opened 2025-06-18
RangeError: Invalid array length


lucasborin:
Hi team,

`cds build --for ord` failed due to an invalid array length. Could you please check it?

Build:
```shell
➜  xxx-foundation git:(main) ✗ cds build --for ord
[ord-plugin] - createAuthConfig: No authorization type is provided. Defaulting to "Open" authentication
building project with {
  versions: { cds: '8.9.4', compiler: '5.0.6', dk: '8.9.5' },
  target: 'gen',
  tasks: [
    { for: 'ord', src: 'srv', dest: 'ord', options: { model: [ '.', 'fts/*', 'srv' ] } }
  ]
}
RangeError: Invalid array length
    at Array.flatMap (<anonymous>)
    at _getEntityTypeMappings (/home/lucasborin/github/xxx-foundation/node_modules/@cap-js/ord/lib/templates.js:371:10)
    at /home/lucasborin/github/xxx-foundation/node_modules/@cap-js/ord/lib/templates.js:265:36
    at Array.forEach (<anonymous>)
    at createAPIResourceTemplate (/home/lucasborin/github/xxx-foundation/node_modules/@cap-js/ord/lib/templates.js:253:11)
    at /home/lucasborin/github/xxx-foundation/node_modules/@cap-js/ord/lib/ord.js:192:9
    at Array.flatMap (<anonymous>)
    at _getAPIResources (/home/lucasborin/github/xxx-foundation/node_modules/@cap-js/ord/lib/ord.js:191:39)
    at module.exports (/home/lucasborin/github/xxx-foundation/node_modules/@cap-js/ord/lib/ord.js:293:30)
    at OrdBuildPlugin.build (/home/lucasborin/github/xxx-foundation/node_modules/@cap-js/ord/lib/build.js:69:29) {
  messages: []
}
```

Code:
![Image](https://github.com/user-attachments/assets/80f9989d-3833-437c-a5bf-d306e4aeff0d)

CDS version:
```shell
➜  xxx-foundation git:(main) ✗ cds version

@cap-js/asyncapi: 1.0.3
@cap-js/cds-types: 0.10.0
@cap-js/change-tracking: 1.0.8
@cap-js/db-service: 1.19.1
@cap-js/hana: 1.8.1
@cap-js/openapi: 1.2.3
@cap-js/ord: 1.3.2
@cap-js/sqlite: 1.10.0
@sap/cds: 8.9.4
@sap/cds-compiler: 5.0.6
[...]

Issue #181, opened 2025-06-23
Slow build process


swennemers:
The `cds build -4 ord` take 90 sec for a middle size project, others (#176 ) talk about 15min. 

Let's profile the build process and check optimization potential.

PR #187, opened 2025-07-07
Fix: error "Invalid array length"

MariusPerleSAP:
Error "Invalid array length" occurred since Array.flatMap tryed to create an array larger than 2^32 - 1

This error can occur in large applications. See #176


swennemers:
it might already cover #181 . Got a big improvement from 260min down to 1:30min for this large app

swennemers:
we plan a patch release tomorrow with 2 more fixes included

Issue #174, opened 2025-06-18
Internal API is sorted into public package


swennemers:
```
{
      "ordId": "sap.foo:apiResource:sap.foo.Product:v1",
      "title": "Product Extraction Service",
      "shortDescription": "Short description of sap.foo.Product",
      "description": "Service to extract product data from the foo product",
      "version": "1.0.0",
      "lastUpdate": "2025-06-17T15:23:34+01:00",
      "visibility": "internal", 
      "partOfPackage": "sap.foo:package:servicecds-api:v1", //<=== should be internal package
```

swennemers:
template.js package assignment done before switching visibility...

PR #195, opened 2025-07-09
Fix DP package assignment

swennemers:
fixing #174


MariusPerleSAP (approved):
So "@DataIntegration.dataProduct.type": "primary" --> visibility internal
👍

PR #196, opened 2025-07-09
Don't publish entity types for SAP policy levels

### LINKED ISSUE #172: Don't publish entity types as SAP application directly
If policy level with sap:core:v1 and sap:dp:v1, don't publish entity types as they are published through global ORD registry. TODO check if it can be done for any policy level that starts with sap...

swennemers:
fix #172

### DIFF
--- a/xmpl/.cdsrc.json
+++ b/xmpl/.cdsrc.json
@@ -13,7 +13,7 @@
     "ord": {
         "namespace": "customer.sample",
         "description": "this is my custom description",
-        "policyLevels": ["none"],
+        "policyLevels": ["sap:core:v1"],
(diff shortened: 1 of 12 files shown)


swennemers:
doesn't work yet

xmpl/.cdsrc.json
```diff
@@ -13,7 +13,7 @@
     "ord": {
         "namespace": "customer.sample",
         "description": "this is my custom description",
-        "policyLevels": ["sap:core:v1"],
+        "policyLevels": ["none"],
```
swennemers, line 16:
```suggestion
        "policyLevels": ["sap:core:v1"],
```

MariusPerleSAP (approved):

PR #212, opened 2025-08-01
Feat: exclude protocol: none and external service

zongqichen:


__tests__/bookshop/package.json
```diff
@@ -15,5 +15,12 @@
     },
     "scripts": {
         "start": "cds-serve"
+    },
+    "cds": {
+        "requires": {
+            "external": {
+                "service": "ExternalService"
+            }
```
swennemers, line 23: Why do you differ from somthing like
````
"cds": {
    "requires": {
        "API_BUSINESS_PARTNER": {
            "kind": "odata-v2",
            "model": "srv/external/API_BUSINESS_PARTNER"
        }
    }
}
````
as in the docs?

File: lib/ord.js [hunk 2]
```diff
@@ -151,8 +150,32 @@ function _shouldSkipIfServiceOnlyContainsEvents(serviceDefinition) {
     return false;
 }
 
+function _shouldNotSkipIfServiceProtocolIsNone(keyDefinition) {
+    if (keyDefinition["_service"] && keyDefinition["_service"]["@protocol"] == "none") {
```
MariusPerleSAP, line 154:
```suggestion
    if (keyDefinition["_service"] && keyDefinition["_service"]["@protocol"] === "none") {
```

MariusPerleSAP (approved):
one line can be perfected otherwise LGTM

Issue #189, opened 2025-07-08
Discuss: Request for default visibility


swennemers:
Some apps request default api resource visibility to be internal, this contradicts previous decisions for CAP services being public interfaces.
They don't want to annotated every service as internal as this is creating too much effort.

Idea: Introduce a configuration flag to toggle the default visibility to internal.

PR #215, opened 2025-08-11
fix: If the visibility is private, do not create a group

KegCoder:
**Tackling** https://github.com/cap-js/ord/issues/189

- Fix: If the visibility is private, do not create a group

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,16 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 The format is based on [Keep a Changelog](http://keepachangelog.com/).
 
+## 1.3.8 (2025-09-01)
+
+## What's Changed
+
+- Feat: Setup Java sample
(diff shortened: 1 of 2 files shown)


Fannon:
Ok, maybe we have a problem in the ORD spec that Group Type and Group do not have a `visibility` and they should have?

For Group Type, I'd expect that you may want to explicitly define the visibility. For groups (instances), probably the same. 

If we generate group (instances), then I'd assume that they inherit the visibility from what they point to. Or actually a bit more complicated: A group that only have private members should be private. If at least one entry is internal, it has to be internal. If there's at least one public member, the group has to be public.

I'll create a PR in the ORD spec to propose this change and discuss it there. Sorry that this will slow down things a bit. 

As an interim solution, we could just not expose groups that only have private members.

Fannon:
Related: https://github.tools.sap/CentralEngineering/open-resource-discovery-specification/pull/1078

This PR would add visibility for group and group type.

zongqichen (approved):
LGTM

PR #224, opened 2025-08-27
Feat: setup java pipeline

zongqichen:


swennemers (approved):
as discussed, it would be good to have a follow up item with a spot test for the ORD config endpoint, the ORD document endpoint and a resource definition file endpoint to return a document, as integration test on the java side.

PR #230, opened 2025-09-02
feat: allow loading of additional attributes for packages

MariusPerleSAP:
There are many more attributes for ORD packages that currently can't be set with .cdsrc and custom.ord.json files.

This change allows passing all values to packages array.

### DIFF
--- a/lib/defaults.js
+++ b/lib/defaults.js
@@ -64,17 +64,10 @@ const mapEnvPackageInfo = (packageConfig) => {
     const filteredEnv = {};
     for (const key in packageConfig) {
         const value = packageConfig[key];
-        if (!value) {
-            continue;
-        }
         const checkFunction = packageTypeChecks[key];
-        if (!checkFunction) {
-            continue;
+        if (value && checkFunction && checkFunction(value)) {
+            filteredEnv[key] = value;
         }
-        if (!checkFunction(value)) {
-            continue;
-        }
-        filteredEnv[key] = value;
     }
     return filteredEnv;
 };


MariusPerleSAP:
Currently: 
no sanitation and no checks --> high performance but usage issue might occur

lib/defaults.js
```diff
@@ -36,6 +36,15 @@ const generateUniquePackageOrdId = (ordNamespace, name, tag, visibility) => {
     return `${ordNamespace}:package:${regexWithRemoval(name)}${tag}${visibilitySuffix}:v1`;
 };
 
+const mapEnvPackageInfo = (appConfigEnv) => {
```
zongqichen, line 39:
```suggestion
const mapEnvPackageInfo = (packageConfig) => {
```
more precisely

lib/defaults.js
```diff
@@ -36,6 +36,15 @@ const generateUniquePackageOrdId = (ordNamespace, name, tag, visibility) => {
...
+    if (!appConfigEnv) {
+        return { vendor: undefined };
+    }
+
```
zongqichen, line 43: I think you are right, we need to consider sanctions, comparing performance, correctness is more important.

So I would suggest:
1. add `ALLOWED_PACKAGE_FIELDS`, maybe move it into constants.js: 
  ```
    const ALLOWED_PACKAGE_FIELDS= new Set([
        'vendor', 'title', 'shortDescription', 'description', 'version',
        'licenseType', 'supportInfo', 'partOfProducts', 'lineOfBusiness', 
        'industry', 'runtimeRestriction', 'tags', 'labels', 
        'documentationLabels', 'packageLinks', 'links', 'policyLevel',
        'customPolicyLevel', 'policyLevels'
    ]);
  ```
2. filter out invalid case:
```
    const filteredEnv = {};
    
    Object.keys(packageConfig).forEach(key => {
        const value = packageConfig[key];
        const valueType = typeof value;
        
        if (ALLOWED_PACKAGE_FIELDS.has(key)) {
            if (Array.isArray(value) || (valueType === 'object' && value !== null)) {
                filteredEnv[key] = value;
            }
            else if (key === 'vendor' && valueType === 'string') {
                filteredEnv[key] = value;
            }
        }
    });
```
Then the invalid case `"packages":["Sales"]`, could be also filtered out

lib/defaults.js
```diff
@@ -36,6 +56,29 @@ const generateUniquePackageOrdId = (ordNamespace, name, tag, visibility) => {
...
+        }
+        const checkFunction = packageTypeChecks[key];
+        if (!checkFunction) {
+            continue;
+        }
+        if (!checkFunction(value)) {
+            continue;
+        }
+        filteredEnv[key] = value;
+    }
```
zongqichen, line 78:
```suggestion
    for (const key in packageConfig) {
        const value = packageConfig[key];
        const checkFunction = packageTypeChecks[key];
        if (value && checkFunction && checkFunction(value)) {
            filteredEnv[key] = value;
        }
    }
```
Lets make it shorter and use positive logic as prefer

zongqichen (approved):
LGTM

PR #264, opened 2025-09-26
feat: add support for @data.product annotation for data product services

swennemers:
- Add DATA_PRODUCT_SIMPLE_ANNOTATION constant for @data.product
- Enhance isPrimaryDataProductService to support both annotations
- @DataIntegration.dataProduct.type: 'primary' takes precedence
- Services with @data.product get full data product ORD properties
- Add comprehensive test coverage with new mock CSN data
- Maintain full backward compatibility

### DIFF
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -44,6 +44,8 @@ const CONTENT_MERGE_KEY = "ordId";
 
 const DATA_PRODUCT_ANNOTATION = "@DataIntegration.dataProduct.type";
 
+const DATA_PRODUCT_SIMPLE_ANNOTATION = "@data.product";
+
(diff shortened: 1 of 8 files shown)


lib/constants.js
```diff
@@ -44,6 +44,8 @@ const CONTENT_MERGE_KEY = "ordId";
 
 const DATA_PRODUCT_ANNOTATION = "@DataIntegration.dataProduct.type";
 
+const DATA_PRODUCT_SIMPLE_ANNOTATION = "@data.product";
```
zongqichen, line 47: Should be `@Data.product`? I would also suggest update docs or create a new docs or README for the change

swennemers: this is how CAP team wants it.

PR #299, opened 2025-11-19
Feat: support mtls

zongqichen:


lib/auth/cf-mtls.js
```diff
@@ -0,0 +1,505 @@
...
+ * Splits by comma and trims whitespace from each token.
+ *
+ * Example:
+ *   "CN=test, O=SAP SE, C=DE"
+ *   → ["CN=test", "O=SAP SE", "C=DE"]
+ *
+ * @param {string} dn - DN-style string
+ * @returns {string[]} Array of DN tokens
+ */
+function tokenizeDn(dn) {
```
jurijat: This needs to be adapted to support also slashes `/`, as UCL will send you the headers in slash-separated. Something like this:
``` 
export function tokenizeDn(dn: string): string[] {
  // Remove leading slash if present
  const cleanDn = dn.startsWith("/") ? dn.substring(1) : dn;
  const separator = dn.startsWith("/") ? "/" : ",";

  return cleanDn
    .split(separator)
    .map((token) => token.trim())
    .filter((token) => token.length > 0);
}
```

Issue #194, opened 2025-07-08
OpenAPI `servers` tag


lucasborin:
Previously, we used to create the OpenAPI artifacts using `cds compile ... --to openapi --openapi:servers "..."`, following the instructions from https://pages.github.tools.sap/cap/docs/advanced/publishing-apis/openapi to inject the servers tag. Now that the ORD library generates these artifacts automatically, I haven't found a way to include this tag. What is the recommended approach?

henrybrink:
Hi @Fannon,

we also face a very similar requirement for consumption of the OpenAPI artifacts in the Business Accelerator Hub. I have opened a small PR which would be sufficient for our requirement and may also fulfil the requirements of @lucasborin. I'm also open for adapting this for further requirements (e.g. setting these options per service and not on a global level). For now, I wanted to keep it as simple as possible, as this is currently blocking us for releasing our documentation.

Best Regards
Henry

PR #314, opened 2025-12-11
Allow custom compiler options for OpenAPI / AsyncAPI

henrybrink:
Hi colleagues,

This PR contains a small change to allow passing compile options to the OpenAPI / AsyncAPI generators. This allows to configure certain options like `openapi:url` / `openapi:servers` which cannot be configured using annotations or enviornment variables. There is already an issue (#194) asking for such an option and we also require it, to properly configure the generated OpenAPI file for consumption in the Business Accellerator Hub. I tried to keep the change as simple as possible, while also not restricting it to servers / url as there may be other options which can be required in certain scenarios.

I'm open for feedback and will adapt the PR if there's something missing from your side / should be changed.

Best Regards
Henry

### DIFF
--- a/__tests__/unit/metaData.test.js
+++ b/__tests__/unit/metaData.test.js
@@ -227,7 +227,7 @@ describe("metaData", () => {
 
     test("getMetadata should pass compile options from cds.env", async () => {
         const url = "/ord/v1/sap.test.cdsrc.sample:apiResource:TestService:v1/TestService.oas3.json";
-        const compileOptions = { "openapi:path": "my_path" };
+        const compileOptions = { "openapi": { "openapi:path": "my_path" } };
         cds.env["ord"] = { compileOptions };
 
         openapi.mockImplementation((csn, options) => {


zongqichen:
Hi @Fannon and @swennemers,

one question about the configuration shape of `cds.env.ord.compileOptions`.

Right now we treat `compileOptions` as a flat bag of options that is passed directly to the OpenAPI compiler. I’m not yet sure how often this will be used in the future, and whether other compilers (besides OpenAPI) might also need their own extra settings.

Would it make sense to already partition this a bit by compiler, e.g.:

```jsonc
"ord": {
  "compileOptions": {
    "openapi": {
      "openapi:url": "https://api.example.com/${service-path}"
    },
    "csn": {
      // future options for CSN compiler etc.
    }
  }
}
```

swennemers:
> Hi @Fannon and @swennemers,
> 
> one question about the configuration shape of `cds.env.ord.compileOptions`.
> 
> Right now we treat `compileOptions` as a flat bag of options that is passed directly to the OpenAPI compiler. I’m not yet sure how often this will be used in the future, and whether other compilers (besides OpenAPI) might also need their own extra settings.
> 
> Would it make sense to already partition this a bit by compiler, e.g.:
> 
> ```json-comments
> "ord": {
>   "compileOptions": {
>     "openapi": {
>       "openapi:url": "https://api.example.com/${service-path}"
>     },
>     "csn": {
>       // future options for CSN compiler etc.
>     }
>   }
> }
> ```

@zongqichen, good idea, yes, the compiler options are different depending on the compilation target, however the options between openapi and asyncapi seem to be shared...

PR #324, opened 2026-01-08
Fix: use @sap/cds instead of @sap/cds-dk for build plugin

zongqichen:
- Change build plugin to extend cds.build.Plugin from @sap/cds
- Remove @sap/cds-dk from peerDependencies
- Update tests to mock cds.build.Plugin before requiring build.js
- Aligns with other CAP plugins (notifications, captivator) pattern
- cds.build.Plugin is injected by cds-dk at build time

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -19,6 +19,7 @@ jobs:
 
             - name: Install dependencies
               run: |
+                  npm install -g @sap/cds-dk@latest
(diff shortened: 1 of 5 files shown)


jiangxin0503:
Hi @zongqichen ,
Can you please release the patch today  if possible, since we need to migrate the CVE issue as soon as possible. 
Many Thanks,
Xin

swennemers (approved):

PR #391, opened 2026-03-12
feat: support tenant-aware ORD for extensible multitenant apps

### LINKED ISSUE #389: Support tenant-aware ORD for extensible multitenant CAP applications
## Problem

When a multitenant CAP application enables extensibility (`cds.requires.extensibility`) or feature toggles (`cds.requires.toggles`), tenants can customize their CDS model via `cds push` — adding entities, fields, and services. Per the ORD specification, these system-instance-aware differences **MUST** be reflected in the ORD document.

Currently, `@cap-js/ord` always generates ORD from the global base model, ignoring all tenant-specific model changes.

### Affected tenant scenarios

| Tenant action | Expected ORD behavior | Current |
|---|---|---|
| `extend Orders with { x_field: String }` | OpenAPI/EDMX includes new field | ❌ base model only |
| Add new `entity x_SalesRegion` | New entityType in ORD | ❌ |
| `extend service with { entity x_Foo }` | New API resource in ORD | ❌ |
| Feature toggle enables/disables a service | Service appears/disappears in ORD | ❌ |

Note: for pure multitenancy without extensibility/toggles, the base ORD document is correctly shared across all tenants — no change needed.

### Root cause

ORD endpoints bypass the CDS middleware stack (`auth` → `ctx-model`), so tenant-specific CSN is never resolved. Other CDS plugins (e.g. `@cap-js/graphql`) correctly integrate with `cds.middlewares`, giving them automatic tenant awareness.

zongqichen:
## Summary

- ORD routes now go through `cds.middlewares.before/after` so that `cds.context.tenant` and `cds.context.model` are populated per request
- When a tenant has extensions (via `cds push`) or feature toggles, the ORD document and metadata (OpenAPI, EDMX) reflect the tenant-specific model
- Config endpoint (`/.well-known/open-resource-discovery`) remains outside middleware per ORD spec

Closes #389

### DIFF
--- a/__tests__/unit/mtx-tenant-aware.test.js
+++ b/__tests__/unit/mtx-tenant-aware.test.js
@@ -1,0 +1,128 @@
+/**
+ * Tests verifying tenant-aware ORD support for extensible multitenant CAP applications.
+ *
+ * When cds.middlewares are used, cds.context.tenant and cds.context.model are populated
+ * per request, allowing tenant-specific CSN (from extensibility/toggles) to be used.
+ */
+const fs = require("fs");
+const path = require("path");
+const cds = require("@sap/cds");
+
+describe("Tenant-Aware ORD", () => {
+    describe("1. metaData.js uses cds.context?.model before service model", () => {
+        test("getMetadata falls back to cds.context.model when no explicit model passed", async () => {
+            const { getMetadata } = require("../../lib/index");
+            const { compile: openapi } = require("@cap-js/openapi");
+
+            // Provide a context model (simulates MTX-populated tenant model)
+            const contextModel = {
+                definitions: {
+                    TestService: {},
+                },
+            };
+            const savedContext = cds.context;
+            cds.context = { model: contextModel };
+
+            jest.mock("@cap-js/openapi", () => ({
+                compile: jest.fn().mockReturnValue("openapi-from-context-model"),
+            }));
+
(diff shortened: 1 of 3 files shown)


mlakov:
@zongqichen - the CDS approach for using the middlewares depends on the idea that tenant & feature toggle information can be extracted from the request. The tenant is by default extracted from the auth information (e.g. from a provided JWT). Feature toggles are by default taken from a specific HTTP header - `Features`. 

Considering the context of UCL/UMS requests - there's a clear gap. For a start - tenant information is provided via an HTTP header `Local-Tenant-Id`. This means that it will not be picked up by the default middlewares, thus leaving the property `cds.context.tenant` empty. The bigger problem is, however, with the feature toggles - requests coming from UCL/UMS have no notion of these at all. The final result of this - ORD metadata will be provided based on CDS models that are not generated with the correct tenant/toggles information.

What is important in this case - we need to consider that as of today CAP for Node.js does not provide a standard API for looking up feature toggles based on a given tenant similar to the [FeatureTogglesInfoProvider](https://javadoc.io/doc/com.sap.cds/cds-services-api/latest/com/sap/cds/services/runtime/FeatureTogglesInfoProvider.html) that exists for CAP for Java (see: https://cap.cloud.sap/docs/guides/extensibility/feature-toggles#in-production).

PR #412, opened 2026-04-28
feat: ODMPLANNING-1151 - Introduce 'perspective: system-version' for static files

### LINKED ISSUE #185: Support ORD 1.12 - static perspective

mlakov:
# Introduce `perspective: system-version` for Static Files and ORD Service

### New Feature

✨ Adds `perspective: "system-version"` and `describedSystemVersion` fields to the ORD document output for both static file builds and the ORD service's runtime document endpoint. The `describedSystemVersion` is resolved from `cds.env["ord"]?.describedSystemVersion` if configured, or falls back to the `version` field in `package.json`. This ensures generated ORD artifacts are correctly tagged with the system-version perspective, enabling downstream consumers to properly identify and handle scoped resources.

Additionally, the `baseTemplate` helper now includes `perspective: "system-version"` in the ORD document entry, and custom ORD content merging is now restricted to a defined allow-list of properties, removing the previous primitive-property warning logic.

### Changes

* `lib/constants.js`: Added `DOCUMENT_PERSPECTIVES` constant with `SystemVersion` and `SystemInstance` values.
* `lib/defaults.js`: Added `resolveDocumentPerspectiveExtension()` helper that returns `perspective` and `describedSystemVersion` (from `cds.env.ord` config or `package.json`). Added `perspective: "system-version"` to the `baseTemplate` document entry.
* `lib/build.js`: Updated `_postProcess` to apply `perspective` and `describedSystemVersion` to the generated ORD document using the new `resolveDocumentPerspectiveExtension()` helper.
* `lib/services/ord-service.js`: Updated the ORD document endpoint to merge perspective extension into the response via `resolveDocumentPerspectiveExtension()`.
* `lib/extend-ord-with-custom.js`: Refactored `compareAndHandleCustomORDContentWithExistingContent` to use an explicit `CUSTOMIZABLE_PROPERTIES` allow-list with `Object.keys().filter().forEach()`, replacing the old `for...in` loop and removing primitive-property warning logic.
* `__tests__/unit/build.test.js`: Added assertions verifying `_postProcess` sets `perspective` to `"system-version"` and populates `describedSystemVersion` from config.
* `__tests__/unit/defaults.test.js`: Added tests for `resolveDocumentPerspectiveExtension()` covering both config-driven and `package.json` fallback scenarios.
* `__tests__/unit/extend-ord-with-custom.test.js`: Removed `warningSpy` and related warning assertions for primitive properties.
* `__tests__/unit/utils/testCustomORDContentFileThrowErrors.json`: Updated test fixture to use known ORD top-level string properties instead of boolean/number primitives.
* `__tests__/integration/cds-build.test.js`: Added snapshot tests and assertions verifying `perspective` and `describedSystemVersion` are present in generated ORD documents; added `fixLastUpdateForStableTest` helper for deterministic snapshots.
* `__tests__/integration/integration-test-app/package.json`: Added `"version": "0.0.1"` to support `describedSystemVersion` derivation from `package.json`.
* `__tests__/integration/integration-test-app/.cdsrc.integrationDep.json`: Added `describedSystemVersion` config entry to test config-driven version resolution.
* `__tests__/integration/__snapshots__/cds-build.test.js.snap`: New snapshot file capturing full ORD document output including `perspective` and `describedSystemVersion`.
* `__tests__/unit/__snapshots__/defaults.test.js.snap`: Updated snapshot to include `perspective: "system-version"` in the `baseTemplate` output.

### Jira Issues

* ODMPLANNING-1151

<details>

**Version:** `1.20.37`

- Correlation ID: `8de144e9-7bdd-4fef-82dc-dde19d597605`
</details>

### DIFF
--- a/__tests__/integration/__snapshots__/cds-build.test.js.snap
+++ b/__tests__/integration/__snapshots__/cds-build.test.js.snap
@@ -1,0 +1,639 @@
+// Jest Snapshot v1, https://jestjs.io/docs/snapshot-testing
+
+exports[`ORD Build Integration Tests IntegrationDependency with cdsrc customization should match snapshot 1`] = `
+{
+  "$schema": "https://open-resource-discovery.github.io/specification/spec-v1/interfaces/Document.schema.json",
+  "apiResources": [
+    {
+      "apiProtocol": "odata-v4",
+      "description": "Description for TestService",
+      "entryPoints": [
+        "/test",
+      ],
+      "extensible": {
+        "supported": "no",
+      },
+      "lastUpdate": "2026-05-04T13:45:01+01:00",
+      "ordId": "customer.capireordintegrationtest:apiResource:TestService:v1",
+      "partOfGroups": [
+        "sap.cds:service:customer.capireordintegrationtest:TestService",
+      ],
+      "partOfPackage": "customer.capireordintegrationtest:package:capireordintegrationtest:v1",
+      "releaseStatus": "active",
+      "resourceDefinitions": [
+        {
+          "accessStrategies": [
+            {
+              "type": "open",
+            },
+          ],
+          "mediaType": "application/json",
+          "type": "openapi-v3",
+          "url": "customer.capireordintegrationtest_apiResource_TestService_v1/TestService.oas3.json",
+        },
+        {
+          "accessStrategies": [
+            {
+              "type": "open",
+            },
+          ],
+          "mediaType": "application/xml",
+          "type": "edmx",
+          "url": "customer.capireordintegrationtest_apiResource_TestService_v1/TestService.edmx",
+        },
+      ],
+      "shortDescription": "Minimal service for ORD integration tests",
+      "title": "Test Service for Integration Testing",
+      "version": "1.0.0",
+      "visibility": "public",
+    },
+    {
+      "apiProtocol": "odata-v4",
+      "description": "Description for SupplierService",
+      "entryPoints": [
+        "/odata/v4/supplier",
+      ],
+      "extensible": {
+        "supported": "no",
+      },
+      "lastUpdate": "2026-05-04T13:45:01+01:00",
+      "ordId": "customer.capireordintegrationtest:apiResource:SupplierService:v1",
+      "partOfGroups": [
+        "sap.cds:service:customer.capireordintegrationtest:SupplierService",
+      ],
+      "partOfPackage": "customer.capireordintegrationtest:package:capireordintegrationtest:v1",
+      "releaseStatus": "active",
+      "resourceDefinitions": [
+        {
+          "accessStrategies": [
+            {
+              "type": "open",
+            },
+          ],
+          "mediaType": "application/json",
+          "type": "openapi-v3",
+          "url": "customer.capireordintegrationtest_apiResource_SupplierService_v1/SupplierService.oas3.json",
+        },
+        {
+          "accessStrategies": [
+            {
+              "type": "open",
+            },
+          ],
+          "mediaType": "application/xml",
+          "type": "edmx",
+          "url": "customer.capireordintegrationtest_apiResource_SupplierService_v1/SupplierService.edmx",
+        },
+      ],
+      "shortDescription": "Short description of SupplierService",
+      "title": "SupplierService",
+      "version": "1.0.0",
+      "visibility": "public",
+    },
+  ],
+  "consumptionBundles": [
+    {
+      "description": "This Consumption Bundle contains all resources of the reference app which are unprotected and do not require authentication",
+      "lastUpdate": "2026-05-04T13:45:01+01:00",
+      "ordId": "capireordintegrationtest:consumptionBundle:noAuth:v1",
+      "shortDescription": "If we have another protected API then it will be another object",
+      "title": "Unprotected resources",
+      "version": "1.0.0",
+    },
+  ],
+  "describedSystemVersion": {
+    "version": "1.0.0",
+  },
+  "description": "this is an application description",
+  "eventResources": [
+    {
+      "description": "CAP Event resource describing events / messages.",
+      "extensible": {
+        "supported": "no",
+      },
+      "lastUpdate": "2026-05-04T13:45:01+01:00",
+      "ordId": "customer.capireordintegrationtest:eventResource:TestService:v1",
+      "partOfGroups": [
+        "sap.cds:service:customer.capireordintegrationtest:TestService",
+      ],
+      "partOfPackage": "customer.capireordintegrationtest:package:capireordintegrationtest:v1",
+      "releaseStatus": "active",
+      "resourceDefinitions": [
+        {
+          "accessStrategies": [
+            {
+              "type": "open",
+            },
+          ],
+          "mediaType": "application/json",
+          "type": "asyncapi-v2",
+          "url": "customer.capireordintegrationtest_eventResource_TestService_v1/TestService.asyncapi2.json",
+        },
+      ],
+      "shortDescription": "Minimal service for ORD integration tests",
+      "title": "Test Service for Integration Testing",
+      "version": "1.0.0",
+      "visibility": "public",
+    },
+  ],
+  "groups": [
+    {
+      "groupId": "sap.cds:service:customer.capireordintegrationtest:TestService",
+      "groupTypeId": "sap.cds:service",
+      "title": "Test Service for Integration Testing",
+    },
+    {
+      "groupId": "sap.cds:service:customer.capireordintegrationtest:SupplierService",
+      "groupTypeId": "sap.cds:service",
+      "title": "Supplier Service",
+    },
+  ],
+  "integrationDependencies": [
+    {
+      "aspects": [
+        {
+          "apiResources": [
+            {
+              "minVersion": "1.0.0",
+              "ordId": "test.sai:apiResource:Supplier:v1",
+            },
+          ],
+          "description": "Integration with Test SAI Supplier Data Product",
+          "mandatory": false,
+          "title": "Test SAI Supplier API",
+        },
+        {
+          "apiResources": [
+            {
+              "minVersion": "1.0.0",
+              "ordId": "test.s4:apiResource:Supplier:v1",
+            },
+          ],
+          "description": "Integration with Test S4 Supplier Data Product",
+          "mandatory": false,
+          "title": "Test S4 Supplier API",
+        },
+      ],
+      "description": "Custom description from cdsrc",
+      "mandatory": false,
+      "ordId": "customer.capireordintegrationtest:integrationDependency:externalDependencies:v1",
+      "partOfPackage": "customer.capireordintegrationtest:package:capireordintegrationtest:v1",
+      "releaseStatus": "beta",
+      "title": "Custom External Dependencies",
+      "version": "2.0.0",
+      "visibility": "internal",
+    },
+  ],
+  "openResourceDiscovery": "1.12",
+  "packages": [
+    {
+      "description": "This package contains public General for capire ord integration test.",
+      "ordId": "customer.capireordintegrationtest:package:capireordintegrationtest:v1",
+      "partOfProducts": [
+        "customer:product:capire.ord.integration.test:",
+      ],
+      "shortDescription": "Package containing public General",
+      "title": "capire ord integration test",
+      "vendor": "customer:vendor:Customer:",
+      "version": "1.0.0",
+    },
+  ],
+  "perspective": "system-version",
+  "policyLevels": [
+    "none",
+  ],
+  "products": [
+    {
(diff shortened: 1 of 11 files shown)


__tests__/unit/build.test.js
```diff
@@ -257,7 +257,13 @@ describe("Build", () => {
                 },
             ],
         };
+
         const updatedOrdDocument = buildClass._postProcess(ordDocument);
+
+        expect(updatedOrdDocument.perspective).toBe("system-version");
+        expect(updatedOrdDocument.describedSystemVersion).toEqual({
+            version: require(path.join(cds.root, "package.json")).version,
```
swennemers: best not to repeat the productive logic in the test

__tests__/integration/cds-build.test.js
```diff
@@ -216,7 +237,7 @@ describe("ORD Build Integration Tests", () => {
             cleanupDir(GEN_DIR);
             buildResult = await runCdsBuild(TEST_APP_ROOT, CDSRC_CONFIG);
             if (fs.existsSync(ORD_DOC_PATH)) {
-                ordDocument = JSON.parse(fs.readFileSync(ORD_DOC_PATH, "utf-8"));
+                ordDocument = resetLastUpdate(JSON.parse(fs.readFileSync(ORD_DOC_PATH, "utf-8")));
```
swennemers: I don't like us manipulating the last update in the tests, this likely let us miss bugs in the calculation. Can't we inject some fixed dates to ease the assertions?

__tests__/unit/build.test.js
swennemers: if we do a mistake there, e.g. cds.root might behave different than expected, we don't catch such issues

__tests__/integration/cds-build.test.js
mlakov: There's no such mechanism as of now. It is also not possible to do it via Jest since these are all deeply nested properties. I can introduce a `.cdsrc.json` override for `lastUpdate` if you're ok with it ?

swennemers: why can't you use jest to stub the time functionality, starting point of the docu: https://jestjs.io/docs/mock-functions#mock-return-values . 

or did you mean you would introduce a custom ord, having lastUpdate overrides? That would also be fine with me, it depends a bit, which parts, we want test. I hope the time bound stuff is tested somewhere in a unit test, so this integration tests don't need to test its details.

mlakov: This is an integration test. It runs in a seprate process  via `npx cds build`. Normall stubbing / mocking via Jest is not applicable whatsoever. The lack of support for custmozing nested property matchin in Jest means that we need to either generate the ORD documents with a stable value for `lastUpdate` or override the generated values for the test (proposed in the current commit).

Considering that some test cases are running on purpose without a `.cdsrc.json` file configured - fixing the `lastUpdate` to a stable value will result in breaking the test logic. Imho the only reasonable approach is to keep the current override of the generated values :(

lib/services/ord-service.js
```diff
@@ -1,11 +1,20 @@
...
 
 const ord = require("../ord.js");
 const Logger = require("../logger.js");
-const defaults = require("../defaults.js");
 const compileMetadata = require("../meta-data.js");
 const { createAuthConfig, createAuthMiddleware } = require("../auth/authentication.js");
 
+const resolveDocumentPerspectiveExtension = () => {
+    return {
+        perspective: "system-version",
```
zongqichen: Extract into constant value?

lib/services/ord-service.js
```diff
@@ -27,13 +36,25 @@ class OpenResourceDiscoveryService extends cds.ApplicationService {
...
+                openResourceDiscoveryV1: {
+                    documents: [
+                        {
+                            url: "/ord/v1/documents/ord-document",
+                            perspective: "system-version",
+                            accessStrategies: authConfig?.accessStrategies || [],
+                        },
+                    ],
+                },
+            });
```
zongqichen: Why do you remove baseTemplate? I can't see the benefit

lib/services/ord-service.js
mlakov: Done.

lib/build.js
```diff
@@ -47,6 +47,11 @@ module.exports = class OrdBuildPlugin extends cds.build.Plugin {
     _postProcess(document) {
         const clone = _.cloneDeep(document);
 
+        clone.perspective = "system-version";
```
zongqichen: I have seen same logic twice

lib/services/ord-service.js
mlakov: `defaults.baseTemplate` is quite the poor name for what it used to be. Removing it makes it simpler to understand what is going on with this API method and what the response of it is.

lib/build.js
mlakov: Yep - once for build-time, once for runtime.

lib/extend-ord-with-custom.js
```diff
@@ -32,23 +32,17 @@ function patchGeneratedOrdResources(destinationObj, sourceObj) {
...
-    for (const key in customORDContent) {
-        const propertyType = typeof customORDContent[key];
-        if (propertyType !== "object" && propertyType !== "array") {
-            Logger.warn(
-                "Found ord top level primitive ord property in customOrdFile:",
-                key,
-                ". Please define it in .cdsrc.json.",
-            );
 
-            continue;
```
zongqichen, line 44: Why do you remove this check and remove the warn? We should not allow user modify the primitive property in customOrd file

__tests__/integration/__snapshots__/cds-build.test.js.snap
```diff
@@ -0,0 +1,639 @@
+// Jest Snapshot v1, https://jestjs.io/docs/snapshot-testing
+
+exports[`ORD Build Integration Tests IntegrationDependency with cdsrc customization should match snapshot 1`] = `
+{
+  "$schema": "https://open-resource-discovery.github.io/specification/spec-v1/interfaces/Document.schema.json",
```
zongqichen, line 5: Why do we need the snapshot for integration test?

mlakov: Why not - being able to see what the expected output is and what the potential delta is - this only makes it easer to support.

lib/extend-ord-with-custom.js
mlakov: This is still not allowed - there's an explicit filter for the properties `$schema`, `namespace` and `openResourceDiscovery`. The behavior is altered in order to allow the modification of the dynamic (per-request) `perspective` property.

lib/services/ord-service.js
zongqichen: If you think it's poor name, you should rename it, instead of removing it. Our previous cleanup and refactoring work is to make the ord-service.js simpler and shorter, not grooming.

zongqichen (changes requested):

__tests__/integration/__snapshots__/cds-build.test.js.snap
zongqichen: If you think is necessary, then please add full snapshot for the rest of integration tests

lib/extend-ord-with-custom.js
zongqichen: Are you sure? Can you test `description` for example?

__tests__/integration/cds-build.test.js
```diff
@@ -8,6 +8,20 @@ const GEN_DIR = path.join(TEST_APP_ROOT, "gen");
 const ORD_GEN_DIR = path.join(GEN_DIR, "srv");
 const ORD_DOC_PATH = path.join(ORD_GEN_DIR, "ord-document.json");
 
+function resetLastUpdate(ord) {
```
swennemers:
```suggestion
function fixLastUpdateForStableTest(ord) {
```

__tests__/integration/cds-build.test.js
swennemers: ok, your call in the end. just suggesting then another function name, to make this patching for the test purpose more obvious.

lib/services/ord-service.js
mlakov: Change has been reverted. Will revisit in another pr.

lib/extend-ord-with-custom.js
mlakov: As stated - `$schema`, `namespace` and `openResourceDiscovery` are excluded. The list does not contain `description`. I can simply add it to the list if you believe it is important.

__tests__/integration/__snapshots__/cds-build.test.js.snap
mlakov: This is going well beyond the scope of the pull request. I'd rather follow up on the topic via a different one.

__tests__/integration/cds-build.test.js
mlakov: Done.

__tests__/integration/cds-build.test.js
mlakov: Renamed.

lib/build.js
```diff
@@ -47,6 +47,11 @@ module.exports = class OrdBuildPlugin extends cds.build.Plugin {
     _postProcess(document) {
         const clone = _.cloneDeep(document);
 
+        clone.perspective = DOCUMENT_PERSPECTIVES.SystemVersion;
+        clone.describedSystemVersion = cds.env["ord"]?.describedSystemVersion ?? {
+            version: require(path.join(cds.root, "package.json")).version,
+        };
```
swennemers: this is duplicate code, you have also in the ord-service-resolveDocumentPerspectiveExtension. can we extract it?

lib/extend-ord-with-custom.js
zongqichen: First of all, it's not the point I believe it is important, lets write code in correct way, you should write testcase to protect this logic, instead of saying `I can simply add it to the list if you believe it is important`. 

Second, this change brings regression bug, this is what I have tested in the second round:
```
#custom.ord.json
{
    "perspective": "non",
    "invalid-key": "hi",
}

#runtime
{
    "perspective": [
        "n"
  ],
    "invalid-key": "hi"
}
```
I think the original logic can cover more corner cases

lib/build.js
mlakov: Done.

lib/extend-ord-with-custom.js
mlakov: The logic is incomplete to begin with - invalid keys that point to an object or an array are not filtered out and will still break the documents. The existing implementation contradicts with the recommended best practice for implementing such constraints via allow/green lists. This is quite evident from the change itself - the reason for having to revisit the said piece of code is only because of the fact that the top-level property `perspective` is of type `string` and not array/object.

It once again boils down to over-inflating the scope of the PR with additional fixes/improvements that go well beyond the originally intended scope. Not that I mind it but.. I do not see this as a good idea, considering the timeline for introducing the changes for supporting perspectives.

mlakov: Allow list has been created. Test case has been updated to include negative coverage for all properties that are not in the allow list + unknown properties.

zongqichen: Hi Martin, I'd appreciate it if you could take a closer look at my comments. The more pressure we have, the more careful we should be. I agree, this change should be out of scope for this PR. However, when I test `perspective` in `custom.ord.json`, the bug still exists. The problem is not the allowlist, it's the merge logic itself. `patchGeneratedOrdResources` uses `_.keyBy(ordId)`, which only works for arrays of ORD resources. When a non-array value (like perspective: "system-version") goes through it, the result is corrupted. That's why we need to filter non-arrays out before calling it.

Also, the unit test case you mentioned is a happy-path test introduced in [this commit](https://github.com/cap-js/ord/commit/590b6c5a7be08940c3e524e19be7f94978aa507b):

```
const result = compareAndHandleCustomORDContentWithExistingContent({}, getCustomORDContent(appConfig));
```
It compares against an empty {}, so key in ordContent is never true and patchGeneratedOrdResources is never called, it won't catch this bug. That's why the filter has been adjusted twice but the underlying issue persists.

mlakov: This is a valid remark but one that leads me to believe that this is a much more serious problem - `groups`, `groupTypes`, `tombstones`, `describedSystemType`, `describedSystemInstance`, `describedSystemVersion`, `policyLevels` - these top-level properties do not provide an `ordId` by default but will pass the check for being array/object.

mlakov: I have completely excluded the usage of the "extension" mechanism for now. Guess the problems with it shall be tracked via a separate issue.

zongqichen (approved):
LGTM, need the following item

PR #413, opened 2026-04-28
feat: ODMPLANNING-1151 - Correctly handle static / tenant-specific metadata

mlakov:
# Correctly Handle Static / Tenant-Specific Metadata for ORD Documents

### New Feature

✨ Introduced tenant-aware model resolution and perspective-based routing for ORD document and metadata endpoints. The service now correctly differentiates between `system-version` (static) and `system-instance` (tenant-specific) perspectives, fetching tenant-specific CDS models via `cds.xt.ModelProviderService` when configured, or falling back to the static global model.

### Changes

* `lib/constants.js`: Added `LOCAL_TENANT_ID_HEADER_KEY` constant (`"local-tenant-id"`) for reading tenant from request headers.

* `lib/defaults.js`: Refactored `baseTemplate` to accept a `tenant` parameter and conditionally include a `system-instance` document link when tenant and toggles/extensibility are configured. Replaced `resolveDocumentPerspectiveExtension` with a new `adjustForPerspective(document, perspective)` method that sets perspective, `describedSystemVersion` (for `system-version`), and appends `?perspective=...` query params to resource definition URLs (for `system-instance`).

* `lib/services/ord-service.js`: Added `validationMiddleware` to validate the `perspective` query param and enforce tenant context for `system-instance` requests. Added `resolveCdsModel(perspective, tenant)` helper to fetch tenant-specific CSN or fall back to `cds.model`. Updated ORD config and document handlers to use tenant/perspective-aware logic. Refactored inline metadata handler into `metadataResponseHandler`. Added `static resolveFeatureToggles(tenant)` method.

* `lib/build.js`: Updated `_postProcess` to use `defaults.adjustForPerspective` with `system-version` perspective, including `perspective` and `describedSystemVersion` fields.

* `lib/ord.js`: Minor cleanup — uses `defaults.$schema` instead of a hardcoded schema string.

* `lib/services/mtx-ord-provider-service.js`: Removed `for` parameter from `getCsn` calls; now wraps result with `defaults.adjustForPerspective` for `system-instance` perspective.

* `__tests__/integration/basic-auth.test.js` & `mtls-auth.test.js`: Expanded integration tests to cover both `system-version` and `system-instance` perspectives, with snapshot-based assertions, query parameter validation, and access strategy verification.

* `__tests__/integration/cds-build.test.js`: Fixed file path resolution using `URL.parse` to handle query params in resource definition URLs.

* `__tests__/integration/integration-test-app/.cdsrc.basic.json` & `.cdsrc.mtls.json`: Added `requires.toggles: true` and `describedSystemVersion` settings.

* `__tests__/integration/integration-test-app/package.json`: Bumped version to `0.1.0` and added `@sap/cds-mtxs` dependency.

* New snapshot files added for `basic-auth` and `mtls-auth` integration test suites; unit test snapshots updated to reflect new `baseTemplate` and `adjustForPerspective` behavior.

### Jira Issues

* ODMPLANNING-1151

<details>

**Version:** `1.20.37`

- Correlation ID: `383eb9f5-a1c7-4fe1-a9fb-f73a252d0c49`
</details>

### DIFF
--- a/__tests__/integration/integration-test-app/.cdsrc.mtls.json
+++ b/__tests__/integration/integration-test-app/.cdsrc.mtls.json
@@ -1,5 +1,11 @@
 {
+    "requires": {
+        "toggles": true
+    },
     "ord": {
+        "describedSystemVersion": {
+            "version": "0.0.1"
+        },
         "authentication": {
             "cfMtls": {
                 "certs": [],
--- a/__tests__/integration/mtls-auth.test.js
+++ b/__tests__/integration/mtls-auth.test.js
@@ -35,6 +35,8 @@ const path = require("path");
 const http = require("http");
 const net = require("net");
 const fs = require("fs");
+
+const utils = require("../utils");
 const { CF_MTLS_HEADERS } = require("../../lib/constants");
 
 const ORD_CONFIG_ENDPOINT = "/.well-known/open-resource-discovery";
@@ -182,7 +184,13 @@ describe("ORD Integration Tests - mTLS Production Mode (cfMtls: true)", () => {
 
         // 1. Create production-style config file: cfMtls: true (requires env var)
         const productionConfig = {
+            requires: {
+                toggles: true,
+            },
             ord: {
+                describedSystemVersion: {
+                    version: "0.0.1"
+                },
                 authentication: {
                     cfMtls: true, // Production mode - requires CF_MTLS_TRUSTED_CERTS
                 },
@@ -289,7 +297,8 @@ describe("ORD Integration Tests - mTLS Production Mode (cfMtls: true)", () => {
 
             const response = await request(BASE_URL).get(ORD_DOCUMENT_ENDPOINT).set(mtlsHeaders).expect(200);
 
-            expect(response.body).toHaveProperty("openResourceDiscovery", "1.14");
+            expect(response.headers["content-type"]).toMatch(/application\/json/);
+            expect(utils.pinLastUpdateForStableTest(response.body)).toMatchSnapshot();
         });
     });
 
@@ -305,13 +314,22 @@ describe("ORD Integration Tests - mTLS Production Mode (cfMtls: true)", () => {
             const response = await request(BASE_URL).get(ORD_DOCUMENT_ENDPOINT).set(mtlsHeaders).expect(200);
 
             expect(response.headers["content-type"]).toMatch(/application\/json/);
-            expect(response.body).toHaveProperty("openResourceDiscovery", "1.14");
+            expect(utils.pinLastUpdateForStableTest(response.body)).toMatchSnapshot();
         });
 
         test("should return ORD config without authentication (endpoint is always open)", async () => {
             const response = await request(BASE_URL).get(ORD_CONFIG_ENDPOINT).expect(200);
 
-            expect(response.body).toHaveProperty("openResourceDiscoveryV1");
+            expect(response.body).toMatchSnapshot();
+        });
+
+        test("should return ORD config for all perspectives without authentication (endpoint is always open)", async () => {
+            const response = await request(BASE_URL)
+                .get(ORD_CONFIG_ENDPOINT)
+                .set("local-tenant-id", "12-34-56")
+                .expect(200);
+
+            expect(response.body).toMatchSnapshot();
         });
 
         test("should accept mTLS headers with DN components in different order", async () => {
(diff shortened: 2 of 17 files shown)


__tests__/integration/integration-test-app/.cdsrc.mtls.json
```diff
@@ -1,5 +1,11 @@
 {
+    "requires": {
+        "toggles": true
+    },
     "ord": {
+        "describedSystemVersion": {
+            "version": "0.0.1"
+        },
```
swennemers, line 8: what about testing the different behavior (extensibility flag + version from package.json), to increase test coverage?

mlakov: This is a great hint. I gave it a try - the `extensibility` flag makes the integration test app too slow though... Not sure whether it is a good idea to use it - makes the tests brittle.

__tests__/integration/mtls-auth.test.js
```diff
@@ -40,6 +40,20 @@ const { CF_MTLS_HEADERS } = require("../../lib/constants");
 const ORD_CONFIG_ENDPOINT = "/.well-known/open-resource-discovery";
 const ORD_DOCUMENT_ENDPOINT = "/ord/v1/documents/ord-document";
 
+function fixLastUpdateForStableTest(ord) {
```
swennemers: we shouldn't repeat the code

__tests__/unit/defaults.test.js
```diff
@@ -180,37 +180,134 @@ describe("defaults", () => {
...
+            expect(defaults.baseTemplate(authConfig, undefined)).toMatchSnapshot();
         });
-    });
 
-    describe("resolveDocumentPerspectiveExtension", () => {
-        it("should return correct value when loading version from .cdsrc.json", () => {
-            cds.env.ord = { describedSystemVersion: { version: "0.1.0" } };
+        it("should return correct value when tenant is not given and only toggles are enabled", () => {
+            cds.env.requires.toggles = true;
+            cds.env.requires.extensibility = false;
```
swennemers, line 184: is this the right env?
https://pages.github.tools.sap/cap/docs/guides/extensibility/customization says to have:
  "cds": {
    "requires": {
      "cds.xt.ExtensibilityService": {

mlakov: As per the official documentation - https://cap.cloud.sap/docs/guides/multitenancy/old-mtx-migration#enable-extensibility - this is the correct way of doing it. Checking against the source code of CAP Node.js confirms this (see: [here](https://github.tools.sap/cap/cds/blob/main/lib/env/schemas/cds-rc.js#L327)).

__tests__/integration/mtls-auth.test.js
mlakov: Done - extracted to `__tests__/utils.js`.

__tests__/unit/defaults.test.js
swennemers: we should check both properties, I also pointed to the official documentation (development version though, but same exist publicly as well), look at the url of your link OLD-MTX-MIGRATION, and cap internally is not only referring to this flag, but also to cds.xt.ExtensibilityService... look a few lines below

mlakov: As far as I get it from the documentation - this is just the way to configure the extensibility service. Additional check with the source code of CAP Node.js (see: [here](https://github.tools.sap/cap/cds/blob/main/lib/srv/middlewares/ctx-model.js#L4)) also suggests that the correct way to verify whether `extensibility` is used or not is via `cds.env.requires.extensibility`. Double checking this against the `.cdsrc.json` schema itself confirms this (see: [here](https://github.tools.sap/cap/cds/blob/main/lib/env/schemas/cds-rc.js#L454)). Same is confirmed when also inspecting the existing properties of `cds.env.requires` itself (see: [here](https://github.tools.sap/cap/cds/blob/a0cb0b924e5bf99e7cdc9f8a668eb734b7312415/lib/env/cds-requires.js#L9))

swennemers (approved):

zongqichen (changes requested):
I need to know why don't we use cap middleware mechanism? All we need to do is mount our routes with cds.app.use("/ord", cds.middlewares.before, router, cds.middlewares.after), it's the common practice as same pattern as @cap-js/graphql and @cap-js/mcp. Then cds.context.model just works, no custom header, no manual ModelProviderService.getCsn() call.

zongqichen (approved):
lgtm

PR #422, opened 2026-05-07
fix: Do not generate OpenAPI resources for ODataV2 services

mlakov:
# Fix: Do Not Generate OpenAPI Resources for OData V2 Services

### Bug Fix

🐛 Fixed an issue where OpenAPI V3 (`openapi-v3`) resource definitions were incorrectly generated for services using the `odata-v2` protocol. OData V2 services only support EDMX format, so the OpenAPI specification entry should not be included in their ORD resource definitions.

### Changes

* `lib/templates.js`: Added a new condition to handle `odata-v2` API protocol separately. When a service is configured with `odata-v2`, only an EDMX resource definition is generated (instead of both OpenAPI V3 and EDMX). Also updated a comment to clarify that the default `else` branch applies specifically to `odata-v4` and other protocols. 

* `__tests__/bookshop/srv/admin-service.cds`: Added `@protocol: 'odata-v2'` annotation to the `AdminService` to reflect the test fixture as an OData V2 service.

* `__tests__/unit/__snapshots__/ord.e2e.test.js.snap`: Updated all affected snapshots to:
  - Change `apiProtocol` from `odata-v4` to `odata-v2` for `AdminService`.
  - Update entry points from `/odata/v4/admin` to `/odata/v2/admin`.
  - Remove the `openapi-v3` resource definition block from `resourceDefinitions` for OData V2 services, leaving only the EDMX definition.

<details>

**Version:** `1.20.43`

- Correlation ID: `375f58f1-89f3-4d26-828e-5af77352976f`
</details>


lib/templates.js
```diff
@@ -363,8 +363,13 @@ const createAPIResourceTemplate = (serviceName, serviceDefinition, appConfig, pa
                         }),
                     },
                 ];
+            } else if (apiProtocol === ORD_API_PROTOCOL.ODATA_V2) {
+                // openapi-v3 is not supported for OData V2, only EDMX
```
swennemers, line 366: @mlakov , can you check this with @Fannon : I remember API guidelines demanding open API also for OData V2 APIs

Fannon: Confirm, @mlakov - we also support OpenAPI v3 (or v2) for OData V2 - at least that's the product standard expectation / BAH expectation. The converter from EDMX to OpenAPI should be able to handle it. Or does it not support it? I guess you had a reason to come to this conclusion

mlakov: @swennemers , @Fannon - The CAP compiler for OpenAPI v3 does not support OData V2. As a result the generated files are empty. I am only assuming that these empty files are actually worse than no files.

Fannon: Ok, I wasn't aware of this. Not sure if this is a really big issue. 
In this case it's better to not publish an empty file

swennemers (approved):

Issue #420, opened 2026-05-07
Ensure @ORD.Extensions.ordId overwrite is applied dynamically


zongqichen:
### Problem

When a user manually specifies an ORD ID via `@ORD.Extensions.ordId` annotation on a service, the top-level `ordId` property is correctly overwritten (via object spread). However, the `resourceDefinitions[].url` is constructed **before** the spread occurs, using the auto-generated `ordId` value:

```js
url: `/ord/v1/${ordId}/${serviceName}.${fileExtension}`
```

This means the resource definition URLs still reference the incorrect auto-generated ORD ID even when a manual override is provided.

### Expected Behavior

When `@ORD.Extensions.ordId` is set, all references to the ORD ID — including resource definition URLs — should reflect the overwritten value.

PR #424, opened 2026-05-08
feat: Support internalNamespace config for ORD ID generation

### LINKED ISSUE #419: Support internal root level namespace for ORD ID generation
### Problem

When an application's internal CAP namespace does not share a prefix with its configured ORD namespace, the generated ORD IDs contain the full internal namespace, producing incorrect and overly verbose IDs.

**Example** (Ariba Sourcing):
- ORD namespace: `sap.sourcing`
- CAP service name: `com.sap.sourcing.api.v1.SourcingMasterDataAPIServiceV1`

**Generated ORD ID**:
```
sap.sourcing:apiResource:com.sap.sourcing.api.v1.SourcingMasterDataAPIServiceV1:v1
```

**Expected** (something like):
```
sap.sourcing:apiResource:SourcingMasterDataAPIServiceV1:v1
```

### Root Cause

`_getGroupNameWithNestedNamespace` removes the ORD namespace prefix from the service name only when the service name starts with it. When the internal namespace (`com.sap.sourcing.api.v1`) diverges from the ORD namespace (`sap.sourcing`), the prefix removal does not apply and the full qualified name ends up in the ORD ID.

Currently there is no way to configure the plugin to handle this mismatch.

swennemers:
When a CAP app's internal CDS namespace differs from its configured ORD namespace, generated ORD IDs previously embedded the full internal namespace prefix. A new optional `cds.env.ord.internalNamespace` setting is now stripped as a fallback when the service name does not start with `ordNamespace`, producing clean IDs like
`sap.sourcing:apiResource:SourcingMasterDataAPIServiceV1:v1` instead of `sap.sourcing:apiResource:com.sap.sourcing.api.v1.SourcingMasterDataAPIServiceV1:v1`.

Also renames `_getGroupNameWithNestedNamespace` → `_getCleanServiceName` for clarity.

Closes #419

### DIFF
--- a/lib/configuration.js
+++ b/lib/configuration.js
@@ -15,6 +15,7 @@ module.exports = class Configuration {
         const packageName = this._loadNameFromPackageJson();
         const eventApplicationNamespace = cds.env?.export?.asyncapi?.applicationNamespace;
         const ordNamespace = cds.env["ord"]?.namespace || `customer.${packageName.replace(/[^a-zA-Z0-9]/g, "")}`;
+        const internalNamespace = env?.internalNamespace;
 
         if (eventApplicationNamespace && ordNamespace !== eventApplicationNamespace) {
             Logger.warn("ORD and AsyncAPI namespaces should be the same.");
@@ -23,6 +24,7 @@ module.exports = class Configuration {
         this._env = env;
         this._packageName = packageName;
         this._ordNamespace = ordNamespace;
+        this._internalNamespace = internalNamespace;
         this._lastUpdate = getRFC3339Date();
         this._serviceNames = this._resolveServiceNames(csn);
         this._existingProductORDId = env?.existingProductORDId;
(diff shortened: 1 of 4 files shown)


lib/configuration.js
```diff
@@ -57,6 +59,10 @@ module.exports = class Configuration {
         return this._ordNamespace;
     }
 
+    get internalNamespace() {
+        return this._internalNamespace;
```
mlakov, line 63: I am only wondering whether it makes sense to allow for more than one internal namespace 🤔

mlakov (approved):
LGTM

zongqichen (approved):
LGTM

lib/configuration.js
swennemers: I would find this dangerous, if we have such mixed namespace situations. in such situation, I would consider #420 as solution and give the app developer the full responsibility over the ORD-ID

PR #428, opened 2026-05-11
fix: Only register well-known compile targets

mlakov:
# Fix: Only Register Well-Known Compile Targets

### Bug Fix

🐛 Resolved an issue where the compile target registration logic was too broad, potentially attempting to register compile targets for plugins that don't support it.

Previously, the filter only excluded `@cap-js/ord` from the registration loop, meaning any other unknown plugin could still be processed. The fix replaces this denylist approach with an allowlist, ensuring only plugins explicitly listed in `PROTOCOL_PROVIDERS` (i.e., `@cap-js/mcp` and `@cap-js/graphql`) are registered as compile targets.

### Changes

* `lib/common/register-compile-targets.js`: Replaced the specific exclusion of `@cap-js/ord` with a positive filter using `PROTOCOL_PROVIDERS`, so only well-known plugins with a defined protocol mapping are processed during compile target registration.

<details>

**Version:** `1.20.43`

- Correlation ID: `44c473e6-3f50-46b7-a269-85a57af8d2f2`
</details>


zongqichen (commented):
I would appreciate new testcase to protect it

zongqichen (approved):
LGTM

PR #431, opened 2026-05-12
chore: Remove generation of default consumption bundle

mlakov:
# Remove Default Consumption Bundle Generation

### Chore

🔧 Removes the automatic generation of a default `consumptionBundles` entry from the ORD document. Previously, a hardcoded "Unprotected resources" consumption bundle was always included in the generated ORD document. This change makes `consumptionBundles` an opt-in field — it is now only included in the output if explicitly provided via the app configuration (`appConfig.env.consumptionBundles`).

### Changes

* `lib/defaults.js`: Removed the `consumptionBundles` default factory function that always generated a static "Unprotected resources" bundle.
* `lib/ord.js`: Changed `consumptionBundles` from an unconditionally included top-level property (falling back to the default) to a conditionally included property that is only added when `appConfig.env.consumptionBundles` is explicitly set and non-empty.
* `__tests__/unit/defaults.test.js`: Removed the unit test for the now-deleted `consumptionBundles` default function.
* `__tests__/unit/split-up-packages-csn.test.js`: Removed the `delete con.lastUpdate` cleanup step that was needed for the default consumption bundle snapshot.
* **Snapshot files** (`basic-auth.test.js.snap`, `cds-build.test.js.snap`, `mtls-auth.test.js.snap`, `defaults.test.js.snap`, `mocked-csn.test.js.snap`, `ord.e2e.test.js.snap`, `split-up-packages-csn.test.js.snap`): Updated all snapshots to remove the previously auto-generated `consumptionBundles` array from expected ORD document outputs.

<details>

**Version:** `1.20.47`

- Correlation ID: `78910b5d-c2d1-4e43-ac48-ea0c77f89aee`
</details>

### DIFF
--- a/docs/ord.md
+++ b/docs/ord.md
@@ -18,7 +18,7 @@
     - [internalNamespace](#internalnamespace)
 7. [ORD Root Properties](#ord-root-property)
 8. [External Data Products](#external-data-products)
-
+9. [Consumption Bundles](#consumption-bundles)
 ---
 
 ## Introduction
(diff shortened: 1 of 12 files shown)


docs/ord.md
```diff
@@ -436,6 +436,40 @@ annotate sap.sai.Supplier.v1 with @ORD.Extensions: {
 
 ---
 
+## Consumption Bundles
+
+Consumption bundles can be configured in one of two ways:
+- Under the `consumptionBundles` property in your `.cdsrc.json` or `package.json` in the `ord` section
```
swennemers:
```suggestion
- Under the `consumptionBundles` property in your `.cdsrc.json` or `package.json` in the `ord` section you define your `consumptionBundle`
```

docs/ord.md
```diff
@@ -436,6 +436,40 @@ annotate sap.sai.Supplier.v1 with @ORD.Extensions: {
...
+- As part of the custom ORD content in your `customOrdContentFile`
```
swennemers: I wouldn't document this as alternative, but as our standard handling that bigger overwrites/own resources are defined in these files.

docs/ord.md
```diff
@@ -436,6 +436,40 @@ annotate sap.sai.Supplier.v1 with @ORD.Extensions: {
 
 ---
 
+## Consumption Bundles
+
+Consumption bundles can be configured in one of two ways:
```
swennemers: I would change this to 1. define the consm. bundle + 2. reference / assign it in the services...

docs/ord.md
mlakov: It does not make much sense to use both options at the same time, doesn't it ?

swennemers (approved):

PR #434, opened 2026-05-12
Feat: compiler pipeline refactoring for dynamic data (ordId)

### LINKED ISSUE #420: Ensure @ORD.Extensions.ordId overwrite is applied dynamically

zongqichen:
`Parse → Resolve → Validate → Emit`
- **Parse**: Collects raw data from the CDS model (services, entities, annotations, protocols) and extensions without making any decisions. Pure data extraction.

- **Resolve**: Constructs ORD resource objects in two steps:
  - Determine each field's final value individually: extensions override defaults for ordId, title, visibility, etc.
  - Once all fields are finalized, derive dependent values (e.g. resourceDefinition paths that embed the final ordId)

- **Validate**: validate value and link are referred correctly

- **Emit**: Pure serialization. Filters empty arrays, merges custom.ord.json, retains only referenced packages. No business logic.


lib/model/visibility.js
```diff
@@ -0,0 +1,40 @@
...
+const {
+    ALLOWED_VISIBILITY,
+    RESOURCE_VISIBILITY,
+    SUPPORTED_IMPLEMENTATIONSTANDARD_VERSIONS,
+    DATA_PRODUCT_ANNOTATION,
+    DATA_PRODUCT_SIMPLE_ANNOTATION,
+    DATA_PRODUCT_TYPE,
+} = require("../constants");
+
+function determineVisibility(extensions, definition, config) {
```
mlakov, line 11: Passing of `extensions` as function argument is not needed:
- `extensions.visibility` is equivalent to `definition["@ORD.Extensions.visibility"]`
- `extensions.implementationStandard` is equivalent to `definition["@ORD.Extensions.implementationStandard"]`

lib/model/visibility.js
```diff
@@ -0,0 +1,40 @@
...
+
+    if (isPrimaryDataProductService(definition)) return RESOURCE_VISIBILITY.internal;
+    if (extensions.visibility) return extensions.visibility;
+    if (definition["@ORD.Extensions.visibility"]) return definition["@ORD.Extensions.visibility"];
+    if (SUPPORTED_IMPLEMENTATIONSTANDARD_VERSIONS.includes(extensions.implementationStandard))
+        return RESOURCE_VISIBILITY.public;
+    return defaultVisibility;
+}
+
+function isPrimaryDataProductService(definition) {
```
mlakov, line 33: The name suggests that the function also checks about `definition.kind === CDS_ELEMENT_KIND.service`.

lib/model/package.js
```diff
@@ -0,0 +1,35 @@
...
+        ...config,
+        policyLevels: config.policyLevels,
+        hasSAPPolicyLevel: hasSAPPolicyLevel(config.policyLevels),
+        existingProductORDId: config.env?.existingProductORDId,
+        products,
+    };
+    return defaults.packages(appConfig, products);
+}
+
+function createConsumptionBundles(config) {
```
mlakov, line 31: Generation of default consumption bundles is to be removed - see PR #431

lib/model/package.js
```diff
@@ -0,0 +1,35 @@
...
+        } else {
+            _.assign(products[0], customProducts);
+        }
+    }
+    config.products = products;
+    return products;
+}
+
+function createPackages(config, products) {
+    const appConfig = {
```
mlakov, line 21: This dynamic tampering with the `appConfig` results in logic that is extremely difficult to follow & debug. This has already been addressed with PR #423 and for me this is a clear regression.

lib/model/package.js
```diff
@@ -0,0 +1,35 @@
...
+    const products = defaults.products(config.packageName);
+    if (config.env?.products) {
+        const customProducts = config.env.products[0];
+        if (customProducts?.ordId?.toLowerCase().startsWith("sap")) {
+            Logger.error("Detected sap product ordId, which should not be defined for custom products.");
+        } else {
+            _.assign(products[0], customProducts);
+        }
+    }
+    config.products = products;
```
mlakov, line 16: This dynamic tampering with the appConfig results in logic that is extremely difficult to follow & debug. This has already been addressed with PR #423 and for me this is a clear regression.

lib/model/naming.js
```diff
@@ -0,0 +1,38 @@
...
+    if (config.internalNamespace && startsWithNamespace(name, config.internalNamespace)) {
+        localName = name.substring(config.internalNamespace.length);
+    } else if (startsWithNamespace(name, config.ordNamespace)) {
+        localName = name.substring(config.ordNamespace.length);
+    }
+    if (localName.startsWith(".")) localName = localName.substring(1);
+    return localName;
+}
+
+function startsWithNamespace(name, namespace) {
```
mlakov, line 16: Isn't this the same as `return name === namespace || name.startsWith(`${namespace}.`);` ?

lib/compiler/index.js
```diff
@@ -0,0 +1,13 @@
...
+const { emit } = require("./emit");
+
+function compile(csn, extensions = []) {
+    const document = parse(csn, extensions);
+    const resolved = resolve(document);
+    validate(resolved);
+    return emit(resolved);
+}
+
+module.exports = { compile, parse, resolve, validate, emit };
```
mlakov, line 13: The name of these functions don't really mean much. E.g. what is `emit` supposed to be doing ?

lib/compiler/index.js
```diff
@@ -0,0 +1,13 @@
+const { parse } = require("./parse");
+const { resolve } = require("./resolve");
+const { validate } = require("./validate");
+const { emit } = require("./emit");
+
+function compile(csn, extensions = []) {
```
mlakov, line 6: From consistency point of view it is weird that all other functions are in dedicated files but `compile is in the `index.js`

PR #438, opened 2026-05-13
fix: Improve generation of entity type ORD IDs when '@ORD.Extensions.version' is specified

mlakov:
# Fix: Improve Generation of Entity Type ORD IDs When `@ORD.Extensions.version` Is Specified

### Bug Fix

🐛 Fixed the generation of ORD IDs for entity types when the `@ORD.Extensions.version` annotation is present. Previously, this annotation was not taken into account when constructing the ORD ID's version segment, causing the version in the ORD ID to remain at `v1` even when a different version was specified.

### Changes

* `lib/templates.js`: Updated `createEntityTypeMappingsItemTemplate` to correctly derive the major version from `@ORD.Extensions.version` (if present) before falling back to the version part of the `@EntityRelationship.entityType` annotation or defaulting to `"1"`. Also fixed the ORD ID construction to always prefix the major version with `v` (e.g., `v2`), and removed a stale TODO comment in `_getEntityVersion`.

* `__tests__/bookshop/db/schema.cds`: Added `@ORD.Extensions.version: '2.0.0'` to the `Authors` entity to cover the new behavior in tests.

* `__tests__/unit/__snapshots__/ord.e2e.test.js.snap`: Updated snapshot expectations to reflect the corrected ORD ID generation — `customer.foo:entityType:Authors:v1` → `customer.foo:entityType:Authors:v2` and entity `version` field updated from `"1.0.0"` to `"2.0.0"`, along with the addition of `@ORD.Extensions.version: "2.0.0"` in the CSN interop content snapshot.

<details>

**Version:** `1.20.47`

- Correlation ID: `42e7c5b0-b519-415e-a5fd-8c7224e78a6d`
</details>


__tests__/bookshop/db/schema.cds
```diff
@@ -18,6 +18,7 @@ entity Books : managed {
         currency : Currency;
 }
 
+@ORD.Extensions.version: '2.0.0'
 @ODM.entityName: 'odm.bookshop.Authors'
 @EntityRelationship.entityType: 'customer.foo:Authors'
```
swennemers: these annotations, if you want to influence the entity type, is inconsistent/wrong.

@EntityRelationship.entityType: 'customer.foo:Authors' indicates a v1 ord id
if you have a v2, you have to mark it as @EntityRelationship.entityType: 'customer.foo:Authors:v2' see https://sap.github.io/csn-interop-specification/spec-v1/extensions/entity-relationship#entity-type

swennemers (changes requested):
not sure about this change and if we should support the version extension here

__tests__/bookshop/db/schema.cds
```diff
@@ -28,6 +29,8 @@ entity Authors : managed {
 }
 
 /** Hierarchically organized Code List for Genres */
+@ORD.Extensions.version: '2.0.0'
```
swennemers:
```suggestion
```
extension version is already tested above, let's have this test case without it

__tests__/unit/__snapshots__/ord.e2e.test.js.snap
```diff
@@ -3416,8 +3544,6 @@ exports[`Tests for products and packages should use valid custom products ordId
   "products": [
     {
       "ordId": "customer:product:eb.bm.tests:",
-      "shortDescription": "Short description of capire bookshop ord sample",
-      "title": "capire bookshop ord sample",
```
swennemers, line 3420: @mlakov the removal of _.assign had side effects...

__tests__/bookshop/db/schema.cds
mlakov: Fixed.

__tests__/unit/__snapshots__/ord.e2e.test.js.snap
mlakov: Done.

swennemers (approved):

PR #440, opened 2026-05-13
fix: Improve generation of integration dependency ORD ID when version is overridden via .cdsrc.json/package.json

mlakov:
# Fix: Correctly Derive Integration Dependency ORD ID from Overridden Version

### Bug Fix

🐛 Fixed an issue where the ORD ID for an integration dependency always used a hardcoded `v1` suffix, even when the version was overridden via `.cdsrc.json` or `package.json`. The major version in the ORD ID is now dynamically derived from the configured version.

### Changes

* `lib/integration-dependency.js`: Extracted the version from `integrationDepConfig.version` (defaulting to `"1.0.0"`), then used the major version component (`v${version.split(".")[0]}`) to construct the `ordId`. Previously, both the `ordId` suffix and the `version` field were hardcoded to `v1` / `"1.0.0"`, ignoring any user-supplied version override.

* `__tests__/integration/__snapshots__/cds-build.test.js.snap`: Updated the snapshot to reflect the corrected ORD ID (`v2` instead of `v1`) when the version is overridden to `2.0.0` via configuration.

<details>

**Version:** `1.20.47`

- Correlation ID: `364cfbd0-0cd6-424c-9069-e6f7882db21d`
</details>

### DIFF
--- a/__tests__/integration/__snapshots__/cds-build.test.js.snap
+++ b/__tests__/integration/__snapshots__/cds-build.test.js.snap
@@ -176,7 +176,7 @@ exports[`ORD Build Integration Tests IntegrationDependency with cdsrc customizat
       ],
       "description": "Custom description from cdsrc",
       "mandatory": false,
-      "ordId": "customer.capireordintegrationtest:integrationDependency:externalDependencies:v1",
+      "ordId": "customer.capireordintegrationtest:integrationDependency:externalDependencies:v2",
       "partOfPackage": "customer.capireordintegrationtest:package:capireordintegrationtest:v1",
       "releaseStatus": "beta",
       "title": "Custom External Dependencies",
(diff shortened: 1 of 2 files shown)


__tests__/integration/__snapshots__/cds-build.test.js.snap
```diff
@@ -176,7 +176,7 @@ exports[`ORD Build Integration Tests IntegrationDependency with cdsrc customizat
       ],
       "description": "Custom description from cdsrc",
       "mandatory": false,
-      "ordId": "customer.capireordintegrationtest:integrationDependency:externalDependencies:v1",
+      "ordId": "customer.capireordintegrationtest:integrationDependency:externalDependencies:v2",
```
swennemers, line 169: good catch to avoid inconsistency

swennemers (approved):

PR #443, opened 2026-05-14
chore: Improve handling of custom ORD content

mlakov:
# Refactor Custom ORD Content Merging with Strategy-Based Approach

### Refactor

♻️ Replaced the generic object traversal merge logic for custom ORD content with an explicit, strategy-based approach using a `MERGE_STRATEGIES` map. This improves correctness, predictability, and eliminates the need for workarounds like `cleanNullProperties`.

### Changes

* `lib/extend-ord-with-custom.js`: Major refactor of the merging logic:
  - Removed `cleanNullProperties` and the old `patchGeneratedOrdResources` functions.
  - Introduced a `MERGE_STRATEGIES` map defining explicit per-property merge behavior (scalar replace, object assign, or smart array merge by key).
  - Added `prune` helper to remove `null`/`undefined` values from merged entities.
  - Added `merge` helper for smart array-of-objects merging using configurable identity keys (e.g., `ordId`, `groupId`).
  - `getCustomORDContent` now consistently returns `{}` instead of `undefined` when no custom file is configured or the file is missing.
  - `compareAndHandleCustomORDContentWithExistingContent` now uses the strategy map to handle properties exclusive to either side or present in both.
  - Removed the warning for top-level primitive properties — they are now handled via the strategy map.
  - Removed the `CONTENT_MERGE_KEY` dependency from constants.

* `__tests__/unit/extend-ord-with-custom.test.js`: Updated test descriptions and expectations to reflect `{}` return values instead of `undefined`; removed the deprecated "warn on top-level primitive" test case.

* `__tests__/bookshop/ord/custom.ord.json`: Removed conflicting top-level properties (`namespace`, `openResourceDiscovery`) that were only needed for the now-removed warning test.

* `__tests__/unit/__snapshots__/extend-ord-with-custom.test.js.snap`: Updated snapshots to reflect the removal of the deprecated test case and corrected `localId` values.

* `__tests__/unit/__snapshots__/ord.e2e.test.js.snap`: Updated snapshots to reflect corrected ordering and content of merged packages, including fixed `shortDescription` and `version` values from the improved merge strategy.

<details>

**Version:** `1.20.51`

- Correlation ID: `6c3ac90f-9ddf-4fa7-ae63-ac5685b83b3a`
</details>


lib/extend-ord-with-custom.js
```diff
@@ -1,68 +1,95 @@
...
+            // Process elements found only in 'customORDContent'
+            ...Object.entries(customORDContent)
+                .filter(([key]) => !(key in ordContent))
+                .map(([key, value]) => [key, structuredClone(value)]),
+
+            // Process elements found in both 'ordContent' and 'customORDContent'
+            ...Object.entries(customORDContent)
+                .filter(([key]) => key in ordContent)
+                .filter(([, value]) => ![null, undefined].includes(value))
+                .map(([key, value]) => [key, MERGE_STRATEGIES[key](structuredClone(ordContent[key]), value)]),
```
zongqichen, line 94: I think we need to provide fallback, otherwise any new property which is not in the list, will crash the plugin.

lib/extend-ord-with-custom.js
```diff
@@ -1,68 +1,95 @@
...
-    const srcObj = _.keyBy(sourceObj, CONTENT_MERGE_KEY);
+const SCALAR_TYPES = Object.freeze(["number", "string", "boolean"]);
 
-    for (const ordId in srcObj) {
-        destObj[ordId] =
-            ordId in destObj
-                ? _.assignWith(structuredClone(destObj[ordId]), structuredClone(srcObj[ordId]))
-                : srcObj[ordId];
-    }
+const MERGE_STRATEGIES = Object.freeze({
```
zongqichen, line 9: Do we need to provide the white list? We will break the flexibility.

lib/extend-ord-with-custom.js
mlakov: The merge is now split in 3 separate parts:

- Properties that only exist in the generated document are directly added to the result
- Properties that only exist in the custom document are directly added to the result
- Properties that exist both in the generated and in the custom document are merged

From that point of view - a problem can only arise when we have a property that we begin generating but we don't have a merge strategy for. This, I believe, we can cover with a unit test in order to ensure that we always maintain our listings up-to-date.

lib/extend-ord-with-custom.js
mlakov: As menitoned in my previous comment - the `MERGE_STRATEGIES`are only relevant for properties that are found in both the generated and custom document. Unknown/custom properties are not processed via this piece of logic.

lib/extend-ord-with-custom.js
```diff
@@ -1,68 +1,95 @@
...
 
-function getCustomORDContent(appConfig) {
-    if (!appConfig.env?.customOrdContentFile) return;
-    const pathToCustomORDContent = path.join(cds.root, appConfig.env?.customOrdContentFile);
-    if (!fs.existsSync(pathToCustomORDContent)) {
-        Logger.error("Custom ORD content file not found at", pathToCustomORDContent);
-        return;
-    }
-    return require(pathToCustomORDContent);
+function merge(source, target, keys) {
```
zongqichen: If you reverse the order merge(target, source), you will have a clean snapshot

lib/extend-ord-with-custom.js
zongqichen: @swennemers, need your input, I'm not sure should we have the whitelist.

mlakov: @zongqichen - as mentioned - this is not a whitelist. This is list of merge strategies for all well-known document properties. I've added a dedicated test case to verify this in `__tests__/unit/extend-ord-with-custom.test.js`.

Handling of unknown/custom properties does not change in comparison with the current implementation from `main`.

zongqichen: I see, I get your point.

mlakov: I have also added a test case that checks the version of the ORD schema used in the test and the one we have under `defaults.openResourceDiscovery`. They should have the same major & minor  (e.g. 1.15)

lib/extend-ord-with-custom.js
mlakov: Test added.

lib/extend-ord-with-custom.js
mlakov: Yep, the review is for sure easier when the snapshots haven't changed. The reason for the change was in fact in the `merge` function - `Array.from(new Set([...Object.keys(targets), ...Object.keys(sources)]))` vs `Array.from(new Set([...Object.keys(sources), ...Object.keys(targets)]))`.  This is all due to the fact that JS sets do preserve insertion order. 

Eventually I swapped the args too - the usage was very confusing - the order differed between the invocation of the merge strategy and the merge function itself. Great catch, @zongqichen.

zongqichen (approved):
LGTM

PR #457, opened 2026-05-18
chore: Extract & refactor 'createEntityTypeTemplate'

mlakov:
# Refactor: Extract & Refactor `createEntityTypeTemplate` into Its Own Module

♻️ **Refactor**: The `createEntityTypeTemplate` function has been extracted from `lib/templates.js` into a dedicated `lib/templates/entity-type.js` module with a cleaner, resolver-based design. The function signature has also been simplified by removing the `packageIds` parameter.

### Changes

* `lib/templates/entity-type.js` *(new)*: Introduces a dedicated module for entity type template creation. Implements a frozen `RESOLVERS` object with individual resolver functions for each field (`ordId`, `title`, `level`, `localId`, `version`, `extensible`, `description`, `releaseStatus`, `shortDescription`, `visibility`, `lastUpdate`, `partOfPackage`). The new `createEntityTypeTemplate(appConfig, entity)` no longer requires an explicit `packageIds` array — package IDs are derived dynamically from `appConfig` using the namespace and app name.

* `lib/templates.js`: Removes `createEntityTypeTemplate`, `_getEntityVersion`, the `LEVEL` and `SEM_VERSION_REGEX` constant imports, and the associated exports. These are now owned by the new module.

* `lib/ord.js`: Updates imports to pull `createEntityTypeTemplate` from `lib/templates/entity-type`. Moves the SAP policy level guard and ODM mapping filter out of the template function and into `_getEntityTypes`, which now also filters out private entity types. The `packageIds` argument is no longer passed to `_getEntityTypes`.

* `__tests__/unit/templates/entity-type.test.js` *(new)*: Comprehensive unit tests covering all individual `RESOLVERS` and the `createEntityTypeTemplate` function, including edge cases for version validation, title annotation precedence, visibility defaults, and package ID construction.

* `__tests__/unit/templates.test.js`: Updates `createEntityTypeTemplate` import to the new module path, removes the `packageIds` argument from all call sites, drops the obsolete SAP policy level and private visibility test cases, and switches entity input shape from `{ ordId, entityName }` to `{ "@EntityRelationship.entityType": "..." }`.

* `__tests__/unit/ord-visibility-split.test.js`: Aligns with the new import path and simplified function signature; updates expected `partOfPackage` values to reflect the new dynamic derivation from `appConfig`.

* `__tests__/unit/__snapshots__/templates.test.js.snap`: Updates snapshot `partOfPackage` values to match the new derivation logic (`customer.testNamespace:package:testAppName:v1`).

<details>

**Version:** `1.20.47`

- Correlation ID: `7ce8d801-d1b2-4d45-8b37-e51f04b76648`
</details>


__tests__/unit/__snapshots__/templates.test.js.snap
```diff
@@ -320,7 +320,7 @@ exports[`templates createEntityTypeTemplate should return entity type with incor
   "level": "root-entity",
   "localId": "SomeAribaDummyEntity",
   "ordId": "sap.sm:entityType:SomeAribaDummyEntity:v3b",
-  "partOfPackage": "sap.test.cdsrc.sample:package:test-entityType:v1",
+  "partOfPackage": "customer.testNamespace:package:testAppName:v1",
```
swennemers, line 323: why did packages changed

mlakov: The test configuration itself was not correct (see: [here](https://github.com/cap-js/ord/pull/457/changes#diff-763b08723f694ebf3b452decea6e89c355b7b523f57ea55f3572829863547232L194)). It was using a package with an incorrect namespace (used: `sap.test.cdsrc.sample`, actual: `customer.testNamespace`, see: [here](https://github.com/cap-js/ord/pull/457/changes#diff-763b08723f694ebf3b452decea6e89c355b7b523f57ea55f3572829863547232L149)) and an incorrect package name as per our strategy for generating package names (see: [defaults.packages](https://github.com/cap-js/ord/blob/main/lib/defaults.js#L100)). We only generate entity types if `appConfig.hasSAPPolicyLevel` returns `false`. At the same time if `appConfig.hasSAPPolicyLevel` returns `false` we generate only the following package ID: `<ord-namespace>:package:<app-name>:v1`. The original test setup was not only using an incorrect namespace but also incorrect app name and was incorrectly using the `-entityType` addition, one that can only be present if the `appConfig.hasSAPPolicyLevel` returns `true`.. but then entity types will not be generated at all.

lib/templates/entity-type.js
swennemers, line 6: @mlakov , I like the approach for properties like title, I don't like this approach, when we only have default values and general @ORD.Extensions overwrites, because I like us to stay flexibility with generic (also with non yet specified properties)...

lib/templates/entity-type.js
```diff
@@ -60,11 +60,9 @@ const RESOLVERS = Object.freeze({
         return entity["@ORD.Extensions.lastUpdate"] ?? appConfig.lastUpdate;
     },
     partOfPackage: (entity, appConfig) => {
-        const visibility = RESOLVERS.visibility(entity, appConfig);
         const name = appConfig.appName?.replace(/[^a-zA-Z0-9]/g, "");
-        const suffix = visibility === RESOURCE_VISIBILITY.public ? "" : `-${visibility}`;
 
-        return entity["@ORD.Extensions.partOfPackage"] ?? `${appConfig.ordNamespace}:package:${name}${suffix}:v1`;
+        return entity["@ORD.Extensions.partOfPackage"] ?? `${appConfig.ordNamespace}:package:${name}:v1`;
```
swennemers: to support BAH publishing, you must not mix resources with different visibility inside the same package. Therefore, I assume, we still need this logic.

lib/templates/entity-type.js
mlakov: The current approach has been designed to support inter-dependencies between properties (e.g. `version` depends on the value of `ordId`, `description` depends on the value of `localId` etc). Always going through a resolver function ensures that these inter-dependencies will be handled in a consistent way. It also allows us to resolve such values from the outside - e.g. if we want to check the `visibility` for an entity type so that we can decide whether we want to include it in the list of exposed entity types for a given API resource.

lib/templates/entity-type.js
mlakov: Whenever we're talking about entity types - they're only generated when `appConfig.hasSAPPolicyLevel` returns `false`. From another point of view - if `appConfig.hasSAPPolicyLevel` returns `false` we only generate a single package: `<ord-namespace>:package:<app-name>:v1`. Based on your comment I am assuming that our logic for package generation is incorrect/incomplete and needs to be fixed 🤔

swennemers: appConfig.hasSAPPolicyLevel returns false might change again, depending on a PoC that starts tomorrow... I still like to have uniform package assignment logic for all ORD resources, even if a certain path is not reachable at the moment for a certain type.

as customers are not publishing to BAH and in discussion with Daniel H. we aligned to have only one package, not following the different package requirements from BAH. So yes, we might have a bug here that customer internal entity types would be assigned to the wrong package...

mlakov: @swennemers - any suggestions on how to continue then ?

lib/templates/entity-type.js
swennemers: ok, then let's discuss on individual properties below.

lib/templates/entity-type.js
```diff
@@ -0,0 +1,99 @@
...
+        const ordId = RESOLVERS.ordId(entity);
+        const version = entity["@ORD.Extensions.version"] ?? ordId.split(":").pop().replace("v", "") + ".0.0";
+
+        if (!SEM_VERSION_REGEX.test(version)) {
+            Logger.warn("Entity version", version, "is not a valid semantic version.");
+        }
+
+        return version;
+    },
+    extensible: (entity) => {
```
swennemers: why is this needed?

lib/templates/entity-type.js
```diff
@@ -0,0 +1,99 @@
...
+            supported: entity["@ORD.Extensions.extensible.supported"] ?? "no",
+            ...(entity["@ORD.Extensions.extensible.description"] && {
+                description: entity["@ORD.Extensions.extensible.description"],
+            }),
+        };
+    },
+    description: (entity) => {
+        return entity["@ORD.Extensions.description"] ?? `Description for ${RESOLVERS.localId(entity)}`;
+    },
+    releaseStatus: (entity) => {
```
swennemers: why is this needed?

lib/templates/entity-type.js
```diff
@@ -0,0 +1,99 @@
...
+        return entity["@ORD.Extensions.releaseStatus"] ?? "active";
+    },
+    shortDescription: (entity) => {
```
swennemers: why is this needed?

lib/templates/entity-type.js
```diff
@@ -0,0 +1,99 @@
...
+    visibility: (entity, appConfig) => {
+        return entity["@ORD.Extensions.visibility"] ?? appConfig.env?.defaultVisibility ?? RESOURCE_VISIBILITY.public;
+    },
+    lastUpdate: (entity, appConfig) => {
+        return entity["@ORD.Extensions.lastUpdate"] ?? appConfig.lastUpdate;
+    },
+    partOfPackage: (entity, appConfig) => {
+        const name = appConfig.appName?.replace(/[^a-zA-Z0-9]/g, "");
+
+        return entity["@ORD.Extensions.partOfPackage"] ?? `${appConfig.ordNamespace}:package:${name}:v1`;
```
swennemers: this logic change is inconsistent with the other ORD resources

swennemers (approved):

PR #463, opened 2026-05-20
fix: Make resource definition URLs OS agnostic

mlakov:
# Fix: Make Resource Definition URLs OS Agnostic

### Bug Fix

🐛 Resolved an issue where resource definition URLs were being constructed using `path.join()`, which produced incorrect backslash-separated paths on Windows systems instead of the expected forward-slash URL format.

### Changes

* `lib/build.js`: Replaced the multi-step `path.join()`-based logic in `_createRelativePath()` with a concise regex-based `replace()` chain. The new implementation strips the `/ord/v1/` prefix and replaces colons with underscores in a single pass, guaranteeing forward-slash separators on all operating systems.
* `__tests__/unit/build.test.js`: Removed the `path` module import and updated expected URL values in assertions from `path.join(...)` expressions to plain forward-slash string literals, reflecting the OS-agnostic behavior.
* `__tests__/integration/mtls-auth-dev.test.js`: Cleaned up imports — added the missing `CF_MTLS_HEADERS` import and removed the now-unused `ORD_ACCESS_STRATEGY` import.

<details>

**Version:** `1.21.1`

- Correlation ID: `715dbd26-1a59-40a9-a46e-e74907048f1c`
</details>


lib/build.js
```diff
@@ -58,9 +58,7 @@ module.exports = class OrdBuildPlugin extends cds.build.Plugin {
     }
 
     _createRelativePath(url) {
-        let relative = url.split("/ord/v1").pop();
-        if (relative.startsWith("/")) relative = relative.slice(1);
-        return path.join(...relative.replace(/:/g, "_").split("/"));
+        return url.replace(/^\/ord\/v1\//, "").replace(/:/g, "_");
```
zongqichen, line 61: It's not a blocker, but we need to think about the /ord/v1 is hardcoded, I can't remember how it can be v2.

zongqichen (approved):
LGTM, and verify it on my windows env

lib/build.js
mlakov: Valid point. It probably makes sense to re-do the regexp so that it allows for any version

PR #466, opened 2026-05-21
chore: Extract 'createEventResourceTemplate'

mlakov:
# Refactor: Extract `createEventResourceTemplate` into Dedicated Module

♻️ **Refactor**: Extracted `createEventResourceTemplate` from `lib/templates.js` into a new dedicated module `lib/templates/event-resource.js`, simplifying its function signature and improving code organization. The function now returns a plain object instead of an array, with `accessStrategies` and `visibility` resolved internally.

### Changes

* `lib/templates/event-resource.js` *(new)*: Dedicated module for event resource template creation. Introduces a `RESOLVERS` object pattern (consistent with `entity-type.js`). Resolves `accessStrategies`, `partOfPackage`, and `visibility` internally using `appConfig`. Returns a plain object instead of an array. Private resource filtering is now the caller's responsibility.
* `lib/templates.js`: Removed `createEventResourceTemplate` and its export.
* `lib/ord.js`: Updated `_getEventResources` to use the new module path and explicitly filter out private resources. Removed the local `_createAuthConfig` helper — auth config is now accessed via `appConfig.authConfig`.
* `lib/configuration.js`: Added `authConfig` initialization via `createAuthConfig()` and a new `authConfig` getter, centralizing authentication configuration in the `Configuration` class.
* `lib/common/utils.js`: Refactored from an inline object export to named function declarations. Extracts and exports `flattenEntityGraph`, `resolveVisibility`, `resolveServiceName`, `isPrimaryDataProductService`, `readORDExtensions`, and `getRFC3339Date`.
* `lib/templates/entity-type.js`: Minor cleanup — simplified version destructuring with a default value and reordered property declarations.
* `__tests__/unit/templates/event-resource.test.js` *(new)*: Added unit tests for all `RESOLVERS` and `createEventResourceTemplate`, covering default behavior and annotation overrides.
* `__tests__/unit/templates.test.js`, `__tests__/unit/namespace.test.js`, `__tests__/unit/ord-visibility-split.test.js`: Updated imports, call sites, and assertions to use the new module path and simplified signature (no `packageIds`/`accessStrategies` args). Added `authConfig` to test `appConfig` fixtures. Removed tests for private visibility array behavior.
* `__tests__/unit/__snapshots__/templates.test.js.snap`, `__tests__/unit/__snapshots__/namespace.test.js.snap`: Updated snapshots to reflect the new plain object return type instead of a single-element array.

<details>

**Version:** `1.20.51`

- Correlation ID: `314e9a42-7029-4a22-bea1-53d2ceafbcf5`
</details>

### DIFF
--- a/lib/common/utils.js
+++ b/lib/common/utils.js
@@ -61,7 +61,7 @@ function resolveServiceName(appConfig, { name }) {
             .filter(Boolean)
             .filter((namespace) => name === namespace || name.startsWith(`${namespace}.`))
             .map((namespace) => name.substring(namespace.length))
-            .pop()
+            .shift()
             ?.replace(/^\./, "") ?? name
     );
 }
--- a/lib/common/utils.js
+++ b/lib/common/utils.js
@@ -3,8 +3,12 @@ const {
     DATA_PRODUCT_TYPE,
     DATA_PRODUCT_SIMPLE_ANNOTATION,
     ORD_EXTENSIONS_PREFIX,
+    RESOURCE_VISIBILITY,
+    ALLOWED_VISIBILITY,
+    SUPPORTED_IMPLEMENTATIONSTANDARD_VERSIONS,
 } = require("../constants");
 const _ = require("lodash");
+const Logger = require("../logger");
 
 function getRFC3339Date() {
     const now = new Date();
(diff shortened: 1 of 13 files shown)


lib/templates/event-resource.js
```diff
@@ -0,0 +1,131 @@
...
+        return service["@ORD.Extensions.ordId"] ?? `${namespace}:eventResource:${name}:${version}`;
+    },
+    title: (service, appConfig) => {
+        return (
+            service["@ORD.Extensions.title"] ??
+            service["@title"] ??
+            service["@Common.Label"] ??
+            service["@EndUserText.label"] ??
+            `ODM ${appConfig.appName.replace(/[^a-zA-Z0-9]/g, "")} Events`
+        );
```
swennemers, line 33: can we unify these (and similar) in a follow PR? I see them quite common across ord resource types

swennemers (dismissed):

lib/common/utils.js
```diff
@@ -3,31 +3,93 @@ const {
...
+function resolveServiceName(appConfig, { name }) {
+    return (
+        [
+            appConfig.internalNamespace, //
+            appConfig.ordNamespace, //
+        ]
+            .filter(Boolean)
+            .filter((namespace) => name === namespace || name.startsWith(`${namespace}.`))
+            .map((namespace) => name.substring(namespace.length))
+            .pop()
```
zongqichen: I'm not sure is a bug or not. But there is a regression diff:
```
ordNamespace      = sap.sourcing                  
internalNamespace = sap.sourcing.internal

name              = "sap.sourcing.internal.BillingService"
```

We should expect `internal.BillingService or BillingService`

lib/common/utils.js
```diff
@@ -3,31 +3,93 @@ const {
...
+        return RESOURCE_VISIBILITY.public;
+    }
+
+    return defaultVisibility;
+}
+
+function resolveServiceName(appConfig, { name }) {
+    return (
+        [
+            appConfig.internalNamespace, //
```
zongqichen, line 58: the `//` is not necessary, right?

lib/templates/event-resource.js
mlakov: It makes a lot of sense, yes.

lib/common/utils.js
mlakov: In this case we should expect `BillingService`. The internal namespace takes precedence over the ORD namespace as per the original implementation. In this regard you're 100% correct - `pop` should be replaced with `shift`.

zongqichen: I guess, it means BillingService, right?

zongqichen (dismissed):
LGTM

lib/common/utils.js
mlakov: Copy-paste mistake, yes.

mlakov: Fixed + tests added.

zongqichen (approved):
LGTM

PR #468, opened 2026-05-22
feat: Split ORD documents bigger than 2mb

mlakov:
# Split ORD Documents Larger Than 2MB

### New Feature

✨ Introduced automatic splitting of large ORD documents that exceed 2MB into multiple smaller parts. When an ORD document exceeds the size limit, it is divided into numbered parts accessible via a `?part=<index>` query parameter. The ORD config endpoint now lists individual part URLs (e.g., `/ord/v1/documents/ord-document?part=0`) instead of a single document URL.

### Changes

* `lib/common/slice.js` *(new)*: Implements the `slice` function that breaks an ORD document into chunks under a specified byte limit, preserving core metadata fields (e.g., `$schema`, `perspective`, `policyLevel`) in each slice.
* `lib/defaults.js`: Added a `sizeLimit` constant (2,000,000 bytes), imported `slice`, and updated `baseTemplate` to dynamically generate document entries per part for both `system-version` and `system-instance` perspectives. Function signature updated to accept `document` and `tenantDocument` instead of `tenant`.
* `lib/services/ord-service.js`: Added support for the `?part` query parameter on the document endpoint, returning only the requested slice. Updated the config endpoint to pass resolved CDS models to `baseTemplate`. Imported `slice` from the new module.
* `__tests__/unit/slice.test.js` *(new)*: Comprehensive unit tests for the `slice` function covering documents below/above the limit, multi-array distribution, metadata preservation, and edge cases.
* `__tests__/unit/defaults.test.js`: Added mock for `../../lib/common/slice`, updated all `baseTemplate` test cases to pass proper `model`/`tenantModel` objects and assert correct `slice` call arguments.
* `__tests__/unit/__snapshots__/defaults.test.js.snap`: Updated all snapshot URLs to include `?part=0`, and multi-perspective URLs to `?part=0&perspective=system-instance`.
* `__tests__/integration/__snapshots__/basic-auth.test.js.snap`: Updated expected document URLs to include `?part=0`.
* `__tests__/integration/__snapshots__/mtls-auth.test.js.snap`: Updated expected document URLs to include `?part=0` and `?part=0&perspective=system-instance`.
* `__tests__/integration/basic-auth.test.js`: Updated inline assertion to expect `?part=0` in document URLs.

<details>

**Version:** `1.21.0`

- Correlation ID: `93e934b6-e981-43fa-ae58-8521855ae22d`
</details>


lib/common/slice.js
```diff
@@ -0,0 +1,65 @@
...
+    );
+
+    let slice = { ...ord };
+    let sliceSize = getFileSizeInBytes(slice);
+
+    Object.keys(document)
+        .filter((key) => !ordMainPartKeys.includes(key))
+        .forEach((key) => {
+            let start = 0;
+
```
MariusPerleSAP: either ordMainPartKey must be expanded or some other projection most prevent failure on non array elements

mlakov: I've added a filter that ensure we only split non-empty array elements.

lib/common/slice.js
```diff
@@ -0,0 +1,67 @@
...
+    const slices = [];
+    const ord = Object.fromEntries(
+        Object.entries(document) //
+            .filter(([key, value]) => ordMainPartKeys.includes(key) || !Array.isArray(value) || !value.length),
+    );
+
+    let slice = {};
+    let sliceSize = 2; // getFileSizeInBytes({});
+    const adjustedLimit = limit - getFileSizeInBytes(ord);
```
zongqichen, line 36: I'm thinking a corner case, if user provides a really huge description > limit, the value of adjustedLimit will be negative. But I don't think it will happen.

lib/common/slice.js
```diff
@@ -0,0 +1,67 @@
...
+                    [key]: value.slice(start, i + chunk),
+                });
+
+                if (remaining < additional) {
+                    if (i > 0 || Object.keys(slice).length > 0) {
+                        slices.push(Object.assign(slice, i === 0 ? {} : { [key]: value.slice(start, i) }));
+                    }
+
+                    // prepare for the next slice
+                    start = i;
```
zongqichen, line 55: There is another rarely case, if i=0, remaining < additional, slice is empty,  it won't push, start will be 0. The case happened when `chunk x itemSize > limit`

lib/common/slice.js
mlakov: There're certain edge cases, indeed. But this is the standard algorithm for the split that is introduced from @MariusPerleSAP for handling static metadata.

lib/common/slice.js
```diff
@@ -0,0 +1,67 @@
+const ordMainPartKeys = Object.freeze([
```
zongqichen, line 1: I think we should add pacakges, products or similar properties in, for example, apiResource will refer partOfPackage.

lib/common/slice.js
mlakov: You're absolutely correct. This is why I introduced this check: `i > 0 || Object.keys(slice).length > 0`. These edge cases should be very difficult to reach though. The current algorightm is a compromise between performance and correctness.

lib/common/slice.js
mlakov: Consumption bundles too.

mlakov: Ok, adding to the final listing - `packages`, `products`, `groups`, `groupTypes`, `consumptionBundles`.

lib/services/ord-service.js
```diff
@@ -80,18 +84,36 @@ class OpenResourceDiscoveryService extends cds.ApplicationService {
...
-            return res.status(200).send(defaults.baseTemplate(authConfig, tenant));
+            return res
+                .status(200)
+                .send(
+                    defaults.baseTemplate(
+                        authConfig,
+                        ord(cds.model, extensions),
+                        !tenant || (!toggles && !extensibility)
+                            ? undefined
+                            : ord(resolveCdsModel(DOCUMENT_PERSPECTIVES.SystemInstance, tenant), extensions),
```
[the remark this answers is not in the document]
zongqichen: I think it makes sense, right?

lib/services/ord-service.js
```diff
@@ -80,18 +84,36 @@ class OpenResourceDiscoveryService extends cds.ApplicationService {
...
+            const perspective = req.query.perspective || DOCUMENT_PERSPECTIVES.SystemVersion;
             const model = await resolveCdsModel(perspective, tenant);
             const extensions = Array.from(Object.values(this.extensions));
 
-            return res.status(200).send(defaults.adjustForPerspective(ord(model, extensions), perspective));
+            return res
+                .status(200)
+                .send(
+                    slice(defaults.adjustForPerspective(ord(model, extensions), perspective), defaults.sizeLimit)[part],
+                );
```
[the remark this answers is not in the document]
zongqichen: I think we should protect here as well.

lib/services/ord-service.js
zongqichen: Bug: resolveCdsModel is called without await, passing a Promise instead of a resolved model to ord(), see hyperspace comment

lib/services/ord-service.js
zongqichen: Bug: part parameter is used as an array index without validation, allowing out-of-bounds access or undefined responses, hyperspace comment as well.

lib/services/ord-service.js
mlakov: Ah, that's quite the great catch. I cannot believe I missed it.

lib/services/ord-service.js
mlakov: Check added.

zongqichen (approved):
LGTM

PR #481, opened 2026-06-03
chore: Update to ORD 1.16

mlakov:
# Update ORD Specification Version to 1.16

### Chore

🔧 Bumped the Open Resource Discovery (ORD) specification version from `1.15` to `1.16` across the codebase, and added `baseUrl` support to the generated ORD document structure.

### Changes

* `lib/defaults.js`: Updated the `openResourceDiscovery` default value from `"1.15"` to `"1.16"`.
* `lib/ord.js`: Added `baseUrl` from `appConfig.env.baseUrl` to the generated ORD document structure.
* `lib/extend-ord-with-custom.js`: Added `baseUrl` as a scalar merge strategy so it can be overridden via custom ORD content.
* `package.json`: Updated the `@open-resource-discovery/specification` dev dependency from `1.15.0` to `1.16.0`.
* `__tests__/unit/__snapshots__/defaults.test.js.snap`: Updated snapshot to reflect the new ORD version `"1.16"`.
* `__tests__/unit/__snapshots__/mocked-csn.test.js.snap`: Updated `openResourceDiscovery` snapshots to `"1.16"`.
* `__tests__/unit/__snapshots__/ord.e2e.test.js.snap`: Updated all relevant `openResourceDiscovery` snapshot entries to `"1.16"`.
* `__tests__/integration/__snapshots__/basic-auth.test.js.snap`: Updated integration test snapshots to `"1.16"`.
* `__tests__/integration/__snapshots__/cds-build.test.js.snap`: Updated integration test snapshots to `"1.16"`.
* `__tests__/integration/__snapshots__/mtls-auth-dev.test.js.snap`: Updated integration test snapshots to `"1.16"`.
* `__tests__/integration/__snapshots__/mtls-auth-prod.test.js.snap`: Updated integration test snapshots to `"1.16"`.

<details>

**Version:** `1.22.4`

- Correlation ID: `8f7e36f0-62f5-4c09-ae05-2d571af97812`
</details>

### DIFF
--- a/lib/ord.js
+++ b/lib/ord.js
@@ -48,7 +48,7 @@ module.exports = (csn, extensions) => {
             [...(extensions || []), getCustomORDContent(appConfig)].filter(Boolean), //
             {
                 $schema: defaults.$schema,
-                baseUrl: appConfig.env.baseUrl,
+                baseUrl: appConfig?.env?.baseUrl,
                 groups: createGroups(appConfig),
                 products: createProducts(appConfig),
                 packages: createPackages(appConfig),
--- a/lib/ord.js
+++ b/lib/ord.js
@@ -48,6 +48,7 @@ module.exports = (csn, extensions) => {
             [...(extensions || []), getCustomORDContent(appConfig)].filter(Boolean), //
             {
                 $schema: defaults.$schema,
+                baseUrl: appConfig.env.baseUrl,
                 groups: createGroups(appConfig),
                 products: createProducts(appConfig),
                 packages: createPackages(appConfig),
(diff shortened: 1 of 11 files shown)


lib/ord.js
```diff
@@ -49,6 +49,7 @@ module.exports = (csn, extensions) => {
             {
                 $schema: defaults.$schema,
                 groups: createGroups(appConfig),
+                baseUrl: appConfig?.env?.baseUrl,
```
Fannon, line 53: @mlakov  Not sure we should do it, if it already works's let's keep it as it is. 

But it should be possible to explicitly override / set if needed

Fannon:
```suggestion
```

Fannon: Misunderstanding on my side: This env is specific to the ORD part of the config, so this is already the explicit way to set baseUrl for ORD

Fannon (approved):

PR #493, opened 2026-06-12
fix: Allow overriding document version for interop CSN

mlakov:
# Allow Overriding Document Version for Interop CSN

### Bug Fix

🐛 Fixed the `add_meta_info` function in the interop CSN module to allow overriding the document version via the `@ORD.Extensions.version` annotation on a service definition, falling back to the version derived from the service name if not specified.

### Changes

* `lib/interop-csn.js`: Updated the `add_meta_info` function to read the `@ORD.Extensions.version` annotation from the service definition when setting `csn.meta.document.version`. If the annotation is not present, the version is derived from the service name as before (e.g., `v1` → `1.0.0`). The meta assignment was also refactored using `Object.assign` for cleaner structure. Additionally, several minor code style fixes were applied (consistent use of `const`, semicolons, and quote style).

<details>

**Version:** `1.22.8`

- Correlation ID: `2fe7e864-3ea6-4d67-966a-9e9aa4e3cb9a`
</details>

### DIFF
--- a/lib/interop-csn.js
+++ b/lib/interop-csn.js
@@ -20,7 +20,7 @@ function add_i18n_texts(csn) {
     // get all texts of the app
     const i18n = [...(localize.bundles4(csn) || [])]
         .filter(([locale]) => !!locale)
-        .map(([locale, value]) => [locale.replaceAll('_', '-'), value])  // CSN interop uses '-' as separator
+        .map(([locale, value]) => [locale.replaceAll("_", "-"), value]) // CSN interop uses '-' as separator
         .reduce((all, [locale, value]) => ({ ...all, [locale]: value }), {});
 
     // get all i18n keys referenced in the csn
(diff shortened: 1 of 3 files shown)


lib/interop-csn.js
```diff
@@ -143,27 +143,33 @@ function add_meta_info(csn) {
...
             srv = segments.pop();
             v = m[1];
         }
-        csn.meta.document = { version: `${v}.0.0` };
-        csn.meta.__name = srv;
-        if (segments.length > 0) csn.meta.__namespace = segments.join(".");
+
+        Object.assign(csn.meta, {
+            __name: srv,
+            document: { version: definition["@ORD.Extensions.version"] ?? `${v}.0.0` },
```
stewsk, line 171: Could you add a check that ensures that major version Data Product (-> v) is the same as the major version specified in @ORD.Extensions.version ?

mlakov: Done.

swennemers (approved):

zongqichen (approved):
LGTM

